PR #1, opened 2016-07-06
Update to match upstream API change (lxd.Client)

yobert:
In lxc/lxd@e3a1cb6b0ac5e1cd5daffa801d377fccc673e94b they updated func (c *Client) Init to have another parameter for passing in a list of "devices".  I think we can safely just put a nil here.


sl1pm4t:
LGTM - thanks!

PR #3, opened 2016-11-27
Update logging and error returns in provider

jtopjian:
Hi there,

I'm playing around with the LXD provider and noticed that flow of the provider wasn't returning the correct error.

For example, if there were any errors with the cert, key, or server cert, I would get:

```
Get https://jtdev:8443/1.0: x509: certificate signed by unknown authority
```

Rather than one of the two errors:

```
root@jtdev:95:96:~/infrastructure/terraform/lxd             
# terraform plan                                            
Refreshing Terraform state in-memory prior to plan...       
The refreshed state will be used to calculate this plan, but
will not be persisted to local or remote state storage.     
                                                            
Error refreshing state: 1 error(s) occurred:                
                                                            
* Certificate or key not found:                             
        /root/.config/lxc/client.crt                        
        /root/.config/lxc/client.key                        
                                                            
root@jtdev:96:97:~/infrastructure/terraform/lxd             
# terraform plan                                            
Refreshing Terraform state in-memory prior to plan...       
The refreshed state will be used to calculate this plan, but
will not be persisted to local or remote state storage.     
                                                            
Error refreshing state: 1 error(s) occurred:                
                                                            
* Server certificate not found:                             
        /root/.config/lxc/servercerts/local.crt                                                                     
```

This PR fixes the above as well as adds some debugging output for the config and client structs.

### DIFF
--- a/lxd/provider.go
+++ b/lxd/provider.go
@@ -2,13 +2,14 @@ package lxd
 
 import (
 	"fmt"
+	"log"
+	"os"
+
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 
-	"github.com/gpmgo/gopm/modules/log"
 	"github.com/lxc/lxd"
 	"github.com/lxc/lxd/shared"
-	"os"
 )
 
 type LxdProvider struct {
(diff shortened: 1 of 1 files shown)


sl1pm4t:
Thanks @jtopjian

PR #4, opened 2016-11-28
Testacc updates

jtopjian:
This update touches up the current container acceptance test. I'm looking into some of the To Do items on this provider and this will help with further testing.


jtopjian:
Ah, that's interesting re the Travis tests. Looks like the Makefile overrode Travis's original workflow. That's now been fixed.

sl1pm4t:
LGTM

PR #5, opened 2016-11-29
Adding support for config argument

jtopjian:
This is an extension of #4. If you'd like to commit these, either #4 needs to be merged first and then I will rebase this off of master or this PR can include both and #4 can be closed without merging. I wanted to split them up for smaller PRs as well as show the direction I was taking with #4.

Let me know if you have any questions.

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -1,6 +1,7 @@
 package lxd
 
 import (
+	"log"
 	"time"
 
 	"github.com/hashicorp/terraform/helper/schema"
@@ -81,6 +82,7 @@ func resourceLxdContainerCreate(d *schema.ResourceData, meta interface{}) error 
 
 	name := d.Get("name").(string)
 	ephem := d.Get("ephemeral").(bool)
+	config := resourceLxdContainerConfigMap(d.Get("config"))
 
 	/*
 	 * requested_empty_profiles means user requested empty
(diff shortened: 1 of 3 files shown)


lxd/resource_lxd_container.go
```diff
@@ -96,38 +98,40 @@ func resourceLxdContainerCreate(d *schema.ResourceData, meta interface{}) error
...
-	// Resource didn't create
-	d.SetId("")
+	if err := client.WaitForSuccess(resp.Operation); err != nil {
+		// Container could not be started
+		return err
+	}
 
-	return err
+	d.SetId(name)
+	return resourceLxdContainerRead(d, meta)
```
jtopjian: I probably should have rolled this into a separate PR. The refactor of this flow more closely aligns with other core Terraform providers. It's generally best to wait until the very end to set the ID. If there was an error, it would have been triggered somewhere above and this point should have never been hit.

sl1pm4t:
Hi @jtopjian  - sorry for the delay, I've been a bit pressed for time lately.
I've merged #4 and added you as a collaborator so you can decide how to handle this PR.
Thanks!

sl1pm4t:
Great work btw - this will be a useful addition

jtopjian:
> sorry for the delay

What delay? :)

> I've merged #4 and added you as a collaborator so you can decide how to handle this PR.

Oh awesome!

PR #11, opened 2016-12-03
Change Profiles from Set to List

jtopjian:
This commit changes the profiles attribute from a set to a list. This is
because profiles are applied in a specified order and a list will enforce
that order.

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -34,7 +34,7 @@ func resourceLxdContainer() *schema.Resource {
 			},
 
 			"profiles": &schema.Schema{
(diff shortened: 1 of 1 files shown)


jtopjian:
@sl1pm4t Just a heads up on this one: converting from a Set to List on existing Terraform deployments will not cause Terraform to break, but it might cause profiles to be reapplied in a certain order. 

This is a prereq to enabling profiles to be updated.

I'll leave this one open for now for if you have any thoughts. :)

sl1pm4t:
Hey thanks for the heads up @jtopjian. I agree this is necessary to make sure the profiles are applied in a specific order, so worth the minimal BC pain.

PR #13, opened 2016-12-06
Support for Container Updates

jtopjian:
This commit introduces support for updating container configuration. Right
now only profile updates are supported.

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -50,6 +50,7 @@ func resourceLxdContainer() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  false,
+				ForceNew: true,
 			},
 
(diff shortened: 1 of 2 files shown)


jtopjian:
This turned out harder than I thought. 

[BriefContainerInfo](https://github.com/lxc/lxd/blob/master/shared/container.go#L112) contains the fields that are able to be updated. From that list:

* We don't want to update `name`, since that's the unique ID of the container. 
* I'm not sure updating `config` is a good idea since random increases and decreases of memory, cpu, and such might have very crazy results on containers.
* I wasn't able to successfully change `ephemeral` on a running container.
* `devices` aren't supported yet

So that leaves `profiles`, which was actually quite easy to implement.

I'll work on adding a `device` attribute once this is merged.

sl1pm4t:
LGTM

PR #15, opened 2016-12-07
Container Profile Fix

jtopjian:
After #13, I noticed that although Terraform reported profiles being updated correctly, the changes weren't being reflected in the container.

This fixes the behavior.


jtopjian:
That Travis failure was intentional -- checking to see if it's working :)

PR #16, opened 2016-12-07
Container device

jtopjian:
I'm not 100% happy with this. I might try a version where `device` looks like:

```hcl
device {
  name = "name"
  type = "disk"
  properties {
    foo = "bar"
    baz = "zab"
  }
}
```

Also still needs docs.


jtopjian:
OK, I like this better. @sl1pm4t thoughts?

sl1pm4t:
The second approach better matches used in other providers, and gives the benefit of validation etc, so LGTM. Thanks!

PR #19, opened 2016-12-11
lxd_network resource

jtopjian:
This commit adds an `lxd_network` resource which can be used to create LXD managed networks and tunnels.

```hcl
resource "lxd_network" "eth1" {
  name = "eth1"

  config {
    ipv4.address = "10.150.19.1/24"
    ipv4.nat = "true"
    ipv6.address = "fd42:474b:622d:259d::1/64"
    ipv6.nat = "true"
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,26 +31,14 @@ provider "lxd" {
 **Resource**
 
 ```hcl
-resource "lxd_container" "test1" {
-  name      = "test1"
-  image     = "ubuntu"     # this assumes an image has been cached locally with the alias 'ubuntu'
-                           # e.g.
-                           # lxc image copy images:ubuntu/xenial/amd64 local: --alias=ubuntu
-  profiles  = ["default"]
-  ephemeral = false
+resource "lxd_network" "eth1" {
+  name = "eth1"
 
   config {
-    limits.cpu = 2         # this must be a valid container config setting or LXD will throw a
-                           # "Bad key: foo" error. See reference below.
-  }
-
-  device {
-    name = "shared"
-    type = "disk"          # Type must be one of none, disk, nic, unix-char, unix-block, usb, gpu
-    properties {           # Properties are valid key/value pairs of the device.
-      source = "/tmp"      # See the LXD documentation for further information.
-      path = "/tmp"
-    }
+    ipv4.address = "10.150.19.1/24"
+    ipv4.nat = "true"
+    ipv6.address = "fd42:474b:622d:259d::1/64"
+    ipv6.nat = "true"
   }
 }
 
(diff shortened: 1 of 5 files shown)


jtopjian:
@sl1pm4t Let me know if you have any thoughts on this, including the restructure of the README file.

With containers now easily able to be part of multiple networks, I'm going to look at the best way to expose the network information in the container resource.

sl1pm4t:
Wow this is cool. I had missed this enhancement to LXD. 
I'll give it a try and provide feedback.

sl1pm4t:
@jtopjian - I gave this a try and it works great. Very cool.

Just a couple of issues I found:
- I was still running LXD 2.0.x stream on my Ubuntu 16.04 host. So creating the network resource I got:
```
Error applying plan:

1 error(s) occurred:

* lxd_network.eth1: not implemented
```

- If someone follows the readme example, they end up with two interfaces in the test1 container. One attached to the `default` profiles bridge and one attached to the 'eth1' bridge. In my case the second one didn't get an IP address. I deleted the container and re-created without the `default` profile and it came up with a single interface on the `eth1` bridge, with an IP address.

Thanks!

jtopjian:
Good points.

* Definitely worth handling the `not implemented` error. I'll get on this.

* WRT the README example: should multiple examples be used at this point? Or have them referenced in a separate document? There's no need for someone to create an `lxd_network` by default, but it does provide a lot of cool possibilities.

* WRT the multiple interfaces: ugh. multiple interfaces are a bear with cloud images. I'll see if there's a quick provisioning step that can be done to bring both `eth0` and `eth1` up. It'd be good for general documentation.

jtopjian:
More thoughts on the example configurations:

* The most basic one should assume the user has set up an LXD environment by the book. This includes running `lxd init` which would provision a network and attach it to the `default` profile. In this case, `lxd_container` is the only resource needed to get up and running.

* A basic profile example that shows how to add cpu and memory limits.

* A network example that can either create a parallel network or do the same work that `lxd init` would have done. This network would be attached to a new profile and seen as `eth0` on the container.

* A second network that would be attached as `eth1`.

* A tunneled network to a second LXD host (I still need to do a POC on this one).

sl1pm4t:
Yes, I think multiple examples in the README are the way to go. You're right, the standalone `lxd_container` example is the minimum viable configuration on a fresh install (and works on LXD 2.0.x), and should be the the taster example in our RM.

It's fairly common in the Terraform docs to have a simple resource example, and a second or third that build on the first to add more complexity.

I don't think the 5 examples you mention would overwhelm the RM. At a later stage we could break it out to a GH wiki if it's getting unwieldy.

Also, that tunneled network example would be very cool - can't wait to see that.

jtopjian:
All sounds good! I'll update the README next.

I just pushed a commit to handle the descriptive error message -- let me know what you think.

sl1pm4t:
LGTM

jtopjian:
OK, the README has been updated. Let me know if it's too verbose.

Also, tunneling is pretty awesome. I want to figure out how to do a tunnel mesh without multicast, but that can wait.

jtopjian:
Awesome! Thanks, Matt.

PR #20, opened 2016-12-12
[WIP] Container Network Block

jtopjian:
This is a sketch of an exported `network` block for containers that will detail all network information. This includes multiple networks and multiple addresses per network.

Technically this could supersede the current `ip_address` and `mac_address` attributes, but I think they're still useful in providing an easy-to-remember attribute to pull that information.

Their equivalent here would be:

* `lxd_container.name.network.0.addresses.ipv4.0`
* `lxd_container.name.network.0.mac_address`

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -99,6 +99,36 @@ func resourceLxdContainer() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"network": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"interface": &schema.Schema{
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"addresses.ipv4": &schema.Schema{
+							Type:     schema.TypeList,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							Computed: true,
+						},
+
+						"addresses.ipv6": &schema.Schema{
+							Type:     schema.TypeList,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							Computed: true,
+						},
+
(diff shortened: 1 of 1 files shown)


jtopjian:
The second commit uses a different design:

* `network.eth0.addresses.ipv4.0`
* `network.eth0.mac_address`

I think this is much more intuitive, but perhaps this block isn't useful after all... As seen in the recent Network examples, `dhclient` must be run manually in order to get an IP on the second interface. Because of this, Terraform won't know of the second IP until someone performs a `refresh`.  That's not a good user experience, IMO.

Of course, the user can upload a custom LXC image that has multiple interfaces configured with DHCP. But I think if the user goes to that length to configure interfaces, they'll know of other methods for obtaining the IPs.

So since we can be sure that `eth0` will always have an IP, that circles back to the already existing `ip_address` attribute. However, I use IPv6 extensively, so I'd like to see an `ip6_address` attribute.

I'm going to close this PR and work on the `ip6_address` attribute.

sl1pm4t:
I know when I first implemented the IP address lookup, I was kind of surprised the only way to retrieve it was by interrogating the guest OS directly. Would be nice if there was better integration between LXD + dnsmasq so LXD could pre-assign a known address and / or query dnsmasq for an address, even if it had not received a DHCP request yet.

jtopjian:
Actually, it's possible to assign static IPs (maybe this is also new from 2.3+?) Check out "Static leases and port security" [here](https://www.stgraber.org/2016/10/27/network-management-with-lxd-2-3/). See [here](https://github.com/lxc/lxd/search?utf8=%E2%9C%93&q=networkUpdateStatic) for the gory details of how it's implemented.

The way a container in `lxd_container` is built, static assignment works :)

I think the downside is that static assignment does not work in client tunnels (since the LXD server would be writing to its local `dnsmasq.host` file - boo).

I'm kind of thinking that a managed way of exec'ing commands from the API is the better way to go than traditional SSH.

PR #28, opened 2017-03-23
Fix failure when container no longer exists

yobert:
If you create a container with terraform and then delete it by hand,
terraform plan fails because client.ContainerState returns an error.

This bug is subtle because go allows partial variable shadowing when
using := to assign to multiple variables. (err is already declared as a
named return parameter.)

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -309,6 +309,9 @@ func resourceLxdContainerExists(d *schema.ResourceData, meta interface{}) (exist
 	exists = false
 
 	ct, err := client.ContainerState(name)
+	if err != nil && err.Error() == "not found" {
+		err = nil
+	}
 	if err == nil && ct != nil {
 		exists = true
 	}


jtopjian:
@sl1pm4t  Looks like Travis needs updated for Go 1.8. Terraform core is using `sort.Slice` which requires it.

yobert:
I think that travis build failure is just due to a go 1.8 standard library change.

yobert:
Yeah. I did run the test suite with `go1.8 linux/amd64` and it passed. Just FYI

jtopjian:
@yobert Cool.

The `Exists` attribute is something I don't have experience with. Given the scenario where a container is deleted out of band of Terraform, will this patch preserve the existing state -- meaning, Terraform will continue to think the container exists?

lxd/resource_lxd_container.go
```diff
@@ -309,6 +309,9 @@ func resourceLxdContainerExists(d *schema.ResourceData, meta interface{}) (exist
 	exists = false
 
 	ct, err := client.ContainerState(name)
+	if err != nil && err.Error() == "not found" {
+		err = nil
```
sl1pm4t: The usual terraform behaviour is to clear the resource ID if we detect the resource no longer exists, for example see the Google Compute Instance code: 
https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_compute_instance.go#L334

We should do the same here, otherwise as @jtopjian points out, Terraform will continue to believe the resource exists.

sl1pm4t: Although, on second thought this is in the Exists func making it subtly different to the Google example.

sl1pm4t:
I updated travis to use go 1.8

lxd/resource_lxd_container.go
yobert: Yeah I also doublechecked the AWS provider code and it does something similar.

yobert: The AWS one doesn't even have an exists function, so maybe that's why in there, they do this on Read:
```
func resourceAwsInstanceRead(d *schema.ResourceData, meta interface{}) error {
   conn := meta.(*AWSClient).ec2conn

   resp, err := conn.DescribeInstances(&ec2.DescribeInstancesInput{
      InstanceIds: []*string{aws.String(d.Id())},
   })
   if err != nil {
      // If the instance was not found, return nil so that we can show
      // that the instance is gone.
      if ec2err, ok := err.(awserr.Error); ok && ec2err.Code() == "InvalidInstanceID.NotFound" {
         d.SetId("")
         return nil
      }

      // Some other error, report it
      return err
   }
```

yobert: I just thought, exists is a cleaner place to put it. I think if you define it, terraform will use it, otherwise it goes right to Read and you do the d.SetId thing. I'm cool with that approach if you like it better? I just think the Exists function is also wrong and we'd need the fix in both places.

jtopjian:
I just tried this out and it works perfectly! @yobert Sorry for questioning this PR so much. 

Named returned arguments can be so confusing sometimes. The key here is the `exists` return arg, which is bool. By default, bool is `false` and since it's never set otherwise in the function, the container doesn't exist.

@yobert is correctly catching the error. When first reading this, it's confusing that someone would suppress an error, but the key is all in the return arg. :)

@sl1pm4t This looks good to me for merging.

> The AWS one doesn't even have an exists function,

Yeah, most of the resources I've looked at haven't used `Exists`. This is why it's so new to me. After this, though, I'll have to look at using them regularly.

sl1pm4t:
Merged. Thanks @yobert + @jtopjian

PR #30, opened 2017-03-30
Custom config directory and automatic authentication

jtopjian:
This commit adds two new provider features:

1. Ability to specify a custom config_dir. This is useful for keeping
LXD configurations in a shared directory.

2. Ability to authenticate to an LXD server. This provides the ability
to not have to authenticate to a new LXD server out-of-band of Terraform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ To generate these files and store them in the LXD daemon, follow these [steps](h
 provider "lxd" {
   scheme  = "https"
   address = "10.1.1.8"
+  remote  = "lxd-server"
 }
 ```
 
@@ -234,10 +235,12 @@ _note_: `local` and `remote` accept both IPv4 and IPv6 addresses.
   * `scheme`   - *Optional* - `https` or `unix`. Defaults to `unix`.
   * `port`     - *Optional* - `https` scheme only - The port on which the LXD daemon is listening. Defaults to 8443.
   * `remote`   - *Optional* - Name of the remote LXD as it exists in the local lxc config. Defaults to `local`.
+  * `remote_password` - *Optional* - Password of the remote LXD server.
+  * `config_dir` - *Optional* - Directory path to client LXD configuration and certs. Defaults to `$HOME/.config/lxc`.
 
 ### Resources
 
-The following resources currently exist:
+The following resources are currently available:
 
   * `lxd_container` - Creates and manages a Container
   * `lxd_network` - Creates and manages a Network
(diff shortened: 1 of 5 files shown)


jtopjian:
This PR is pretty awesome, if I do say so :)

It enables someone to set up a new LXD server and provision LXD containers entirely through Terraform. I haven't tried this specific case out yet, but in theory, it would work on a non-Linux workstation, or any workstation without the LXD/LXC packages.

sl1pm4t:
Awesome. I've been wanting to do this for a while too.

jtopjian:
Just FYI that there are two caveats with this implementation:

1. I had to copy some private LXC functions out of the LXD API. This isn't optimal, but nothing else can be done.
2. It only supports PKI right now. I included an example Terraform configuration for setting up a full PKI environment. It's pretty lengthy, so I added it to the `test-fixtures` dir rather than the README.

The `addRemote` function could be extended to support non-PKI environments in the future. It would require auto-accepting a cert, though.

sl1pm4t:
Sorry for the delay @jtopjian - testing and reviewing now.

jtopjian:
No rush at all. The review request wasn't a poke :)

sl1pm4t:
I get the following error when I try connect to an LXD daemon (whose cert was generated by `lxd init` and client cert generated with `lxc remote add ....`):
```
* provider.lxd: Get https://util:8443/1.0: x509: certificate signed by unknown authority
```

Perhaps it's worth adding an insecure mode to the provider config to accept self signed certs?!

jtopjian:
> Perhaps it's worth adding an insecure mode to the provider config to accept self signed certs?!

That's actually done in the [LXD API](https://github.com/lxc/lxd/blob/0259fffff5764817cf0096a80d823dc24989c5e0/lxc/remote.go#L86) I got hung up on that error, too, and it turned out I wasn't connecting to the LXD server with the correct name as written in the cert. I created the cert based on IP and once I tried connecting via IP (instead of hostname or /etc/host name), it let me in.

sl1pm4t:
Actually, you're on the right track @jtopjian . I just updated the provider config and set the correct name for the provider ( `foo` -> `localhost` ). `localhost` was what I used when doing `lxc remote add ....`.
After that change I don't get that error.

jtopjian:
That is good news. I did a cursory look and I don't think the Go http client is easily exposed. :)

sl1pm4t:
Merged. Thanks for this @jtopjian !

jtopjian:
Thanks, Matt!

PR #31, opened 2017-04-02
Container File Support

jtopjian:
This commit adds the file block to the lxd_container resource.
One or more individual files may be defined in the resource and
they will be uploaded/pushed to the container via the LXC/LXD API.

Updates to the files will cause all old files to be deleted and
new copies to be uploaded.

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -3,6 +3,11 @@ package lxd
 import (
 	"fmt"
 	"log"
+	"os"
+	"path"
+	"path/filepath"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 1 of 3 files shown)


lxd/resource_lxd_container.go
```diff
@@ -85,6 +90,44 @@ func resourceLxdContainer() *schema.Resource {
...
+							Optional: true,
+						},
+
+						"gid": &schema.Schema{
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+
+						"mode": &schema.Schema{
+							Type:     schema.TypeString,
```
sl1pm4t: What is the reason for making this a string, rather than Int ?
It seems later on, you expect octal file mode only, so why allow string input?

sl1pm4t:
Other that my one review question, this is working great and is a very useful addition - thanks!

lxd/resource_lxd_container.go
jtopjian: I considered int, but went with string for two reasons:

1. It matches LXC, including parsing to int and such.
2. It enables modes like "0644", even though a zero is added later if missing.

I'm happy to change it to int, though.

sl1pm4t: You're quite right. I also realised the reason some time after posing the question.

PR #32, opened 2017-04-02
Add "refresh_interval" parameter

yobert:
This lets you configure how aggressively the client polls the server for
state changes. If your containers start up very fast, you may want to
decrease this interval for a speed boost.


yobert:
Use case: Almost all my containers start up in about 500ms. So being able to set this to 1s is a very nice thing.

yobert:
What would be even better is to use LXD's blocking API calls. But looking through terraform that looks pretty hard, and I don't think it would play nice with their timeouts...

jtopjian:
Agreed. Terraform recently added support for custom CRUD timeouts, but that doesn't apply to delays, so I think what you're doing is valid.

This gets a +1 from me :)

yobert:
Sorry this is taking me so long. Trying to grok all these test helpers in terraform...

yobert:
Updated the tests to have that configuration value, but it's a little stupid. Sorry, I know this could be better.

sl1pm4t:
Thanks @yobert

PR #34, opened 2017-04-06
Allow image names with remote: prefix

sl1pm4t:
This PR allows a container to be launched using a remote image - without pre-caching the image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,7 @@ resource "lxd_container" "test1" {
 }
 ```
 
-You can also launch a container directly from a remote image, not locally cached, by referencing the remote image name using the format:
- > `[remote:]image_alias|image_hash`
+You can also launch a container directly from a remote image, not locally cached, by referencing the remote image name using the format `[remote:]<image_alias|image_hash>`
 
 ```hcl
 resource "lxd_container" "test1" {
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -49,6 +49,17 @@ resource "lxd_container" "test1" {
 }
 ```
 
+You can also launch a container directly from a remote image, not locally cached, by referencing the remote image name using the format:
```
jtopjian: Small nit, but maybe add a note that the remote must already exist and that the `images` is usually enabled by default. It can be verified with `lxc remote list`.

sl1pm4t: Could you elaborate on 2. 
In my tests the image was cached locally but without any aliases?!

jtopjian: bah. I had too many images and got them mixed up :)

I flushed everything and retried. Aliases aren't set by this feature, so no harm done at all. Sorry about that!

sl1pm4t: Cool thanks. No worries - better to be certain.
Also, you're right this needs more documentation - the remotes that are accessible are only the LXD server itself and the 3 pre-canned LXD defaults (images, ubuntu, ubuntu-daily), which needs explanation.

jtopjian:
:+1: 

LGTM!

PR #37, opened 2017-04-16
Volume resource

jtopjian:
Depends on #36

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,20 @@ With these resources in place, attach them to a profile in the exact same way de
 
 _note_: `local` and `remote` accept both IPv4 and IPv6 addresses.
 
+#### Storage Pools
+
+To create and manage Storage Pools, use the `lxd_storage_pool` resource:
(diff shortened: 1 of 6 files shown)


sl1pm4t:
LGTM - however a couple of questions:
- this PR seems to contain the changes from #36 as well - do you intend to close #36 in favour of this one?
- will you be updating the `lxd_container` resource to make use of storage pools / volumes?

jtopjian:
> this PR seems to contain the changes from #36 as well - do you intend to close #36 in favour of this one?

I can do that, yes. I broke them up into 2 PRs to make review easier, but I guess this one just has all of the changes for both, so not so simple to review in the end :)

> will you be updating the lxd_container resource to make use of storage pools / volumes?

Yes, I'll be making a dedicated `lxd_container_volume_attach` resource to handle attaching and detaching.

sl1pm4t:
Ok Thanks

PR #38, opened 2017-04-16
lxd_snapshot resource

sl1pm4t:
This allows creating, destroying container snapshots.

PR #39, opened 2017-04-17
lxd_volume_container_attach resource

jtopjian:
I can rebase when #38 is merged, if needed.

It's also possible to attach a volume to a profile, so I'll be working on `lxd_volume_profile_attach` next.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 test:
 	go get ./...
 	go get github.com/dustinkirkland/golang-petname
-	go test -v ./lxd
+	go test -timeout 20m -v ./lxd
 
 testacc:
 	TF_LOG=debug TF_ACC=1 go test ./lxd -v $(TESTARGS)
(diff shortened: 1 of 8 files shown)


jtopjian:
There we go. We have too many tests 😄

sl1pm4t:
How does one set file permissions / ownership for attached volumes? 
I've tried creating a volume from `dir` and `zfs` based pool, but root inside the container does not have write permissions.

Also the owner is:
```
drwxr-xr-x  2 nobody nogroup  2 Apr 18 20:36 attach-test
```

jtopjian:
Yeah, this is unfortunately a whole other issue. There are quite a few open and closed issues about this on the LXD repo. I don't have links off-hand, but search for `setfacl`. I think the issue borders on what's under LXD's realm and not which is why there isn't a good solution yet. IMO, LXD should handle setting permissions on new volumes.

I've only tested this with the dir and btrfs backends and had to use `setfacl` for both. I'm sure ZFS has the same problems, but I'm not sure about LVM.

And this isn't specific to the volume/attach resources, either. Same thing happens when you define a device in a profile or container.

PR #40, opened 2017-04-19
LXD Server Authentication Fix

jtopjian:
Poking around the LXD code, I found `client.AddMyCertToServer` which I missed before. Simplifies authentication quite a bit.

Unfortunately downloading the server certificate, from what I can tell, is implemented as a private method.


jtopjian:
Fortunately there was a build error -- I spotted an issue where a client would never be able to authenticate if the password was incorrect the first time. Tried to clean the workflow up, but it's still a bit complicated.

sl1pm4t:
This reminds me of a potential misuse problem I've been pondering.

Every time we use `client.AddMyCertToServer`, the LXD daemon is storing our client certificate in it's DB - so imagine the scenario where someone is running Terraform changes from some CI/CD tool (Travis, Drone or possibly even HashiCorp Atlas) where the "build" environment is ephemeral.
If not configured optimally, the LXD provider will generate a new client certificate on each run, and store in the LXD trust DB, but the cert file is lost after each run. 

I wonder how long before hundreds/thousands of orphaned trust certs start to impact LXD API performance?

Of course, ideally in this scenario the client cert is created using a Terraform resource, and saved in the `.tfstate`, but as an alternative could we use something like is discussed [here](https://github.com/hashicorp/terraform/issues/6258) to delete the auto generated cert on exit?

jtopjian:
Ah, yeah that's a good point. I think it would take _a lot_ of certificate churn to start causing a problem, but having short-lived clients certainly leaves junk around. I guess if abused enough, this opens the LXD server to some type of DOS.

Maybe if LXD kept track of a "last seen" time of the client, the admin could easily purge certificates not seen in x days.

In addition, using PKI doesn't trigger a certificate exchange, so deploying a client with a provisioned cert/key is another option.

As I've discovered, though, LXD's PKI is quite light. For example, it doesn't do any verification of the client certificate. As long as the CA matches, the client is authenticated. So another option is to use an SSL terminator that forwards traffic to the LXD local socket. Now LXD isn't handling certificates at all :)

jtopjian:
re travis failure: what the what? I'll get that fixed.

PR #42, opened 2017-04-22
Cached Image resource

sl1pm4t:
This PR adds the ability to manage cached images. Pull from remote, set aliases etc.

Edit: I've named it `lxd_cached_image` to distinguish it from a future resource I'm planning to create, that will allow users to create an LXD image from a tarball or container snapshot.


sl1pm4t:
Hmmm.. strange.

sl1pm4t:
This PR is getting a bit noisy, so will close and open a replacement.

PR #43, opened 2017-04-23
Add lxd_cached_image resource

sl1pm4t:
This adds the `lxd_cached_image` resource. It can used to manage local
copies of remote images, set aliases etc.

I've named it `lxd_cached_image` to distinguish it from a future
resource I'm planning to create, that will allow users to create an LXD
image from a tarball.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,10 @@ LXD Resource provider for Terraform
 
 This provider connects to the LXD daemon over local Unix socket or HTTPS.
 
-It makes use of the [LXD client library](http://github.com/lxc/lxd), which currently looks in `~/.config/lxc/` for `client.crt` and `client.key` files that must exist to be able to communicate with the LXD daemon.
+It makes use of the [LXD client library](http://github.com/lxc/lxd), which currently looks in `~/.config/lxc/` for `client.crt` and `client.key` files to authenticate against the LXD daemon.
 
-To generate these files and store them in the LXD daemon, follow these [steps](https://linuxcontainers.org/lxd/getting-started-cli/#multiple-hosts).
+To generate these files and store them in the LXD client config, follow these [steps](https://linuxcontainers.org/lxd/getting-started-cli/#multiple-hosts).
+Alternatively, the LXD Terraform provider can generate them on demand if `generate_client_certificates` is set to true.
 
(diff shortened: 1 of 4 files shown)


jtopjian:
Nice resource :)

With regard to `aliases`, `copy_aliases`, and `copied_aliases`, if I'm understanding it correctly, this is a common pattern emerging in some Terraform resources. I recently went through this with two OpenStack resources. The recommended approach was to treat one as an "input" and one as a computed read-only "output". So you could have `aliases` as the input and `all_aliases` as the output. (`all_` being the suggested prefix).

In the `Read` function, you would no longer do a `d.Set("aliases", x)` since the user's configuration will be the source of truth. `all_aliases`, on the other hand, will be the API's source of truth. Depending on the results, the value of `all_aliases` might just equal `aliases`, or a combination of `aliases` and upstream-specified aliases. 

Makes it a lot easier to handle these situations :)

And I think you could remove the `copy_aliases` attribute by doing this, too.

Thoughts?

sl1pm4t:
Yes, the aliases are a bit of a PITA - and will probably go mostly unused. Especially if all the resources are managed with Terraform, there shouldn't be much need for the aliases.
However, having already coded up handling of the aliases I might as well keep it in place. 
If I fall back to a `aliases` input that doesn't get updated on `Read`, then it would make it impossible to ever add or remove aliases right?!

`copied_aliases` is slightly different to `all_aliases`, in that it only includes the aliases that were defined in the remote LXD server at the time the image copy was created. If it were possible, I would make it an internal hidden attribute used only for the purposes of suppressing diffs.

I could also add `all_aliases` output that you mention, that reflects the merged set of aliases.

jtopjian:
OK, I understand. The key here is the ability to opt-in to setting aliases. If it was mandated that downloaded/cached images received the upstream aliases, then the `alias`, `all_aliases` combo would make sense.

Give me a day or so to test this and poke at it. :)

jtopjian:
This looks good to me -- everything does make sense.

I opened #44 which uses Terraform core's built-in set functions, but really it's six one way, half a dozen the other. I think playing around with the scenario of the aliases/copied_aliases can go on all day, so feel free to discard it. :)

PR #44, opened 2017-04-25
Use core set functions

jtopjian:


sl1pm4t:
Wow - I've never seen anyone do a PR to a non-master branch. That's cool.

And re-using the core terraform logic makes sense. 👍

Issue #21, opened 2016-12-12
Thoughts on Profile/Container Configuration


jtopjian:
All LXD resources can take a set of key/value pairs for "configuration". LXD supports a large, though finite, amount of [configuration](https://github.com/lxc/lxd/blob/master/doc/configuration.md).

AFAICT, LXD allows _all_ configuration to be updated in real-time. I haven't made any of the `config` blocks able to be updated because, IMO, updating something like `limits.cpu` while a container is running could be dangerous. 

Thoughts:

* Maybe this provider should yield to the LXD API and update anything LXD wants to update. Discretion is left to the user.
* What are the benefits of creating resources that more closely model the available configuration options? For example:

```hcl
config {
  boot {
    autostart = true
  }

  limits {
    cpu = 2
  }
}
```

I'm not a huge fan of this because keys are stored within LXD as `limits.cpu` verbatim. I think doing this would be overmodeling things.

* If this provider should play it safe and not update certain configuration options, what about the keys that are described as [user defined](https://github.com/lxc/lxd/blob/0c63f15612ea6bdc59a0898e2d124df4d34532e3/shared/container.go#L344-L345)? for example:

```hcl
config {
  limits.cpu = 2
}

# user_data.role would translate to config["user.role"]
user_data {
  role = "mysql"
}

environment {
  FOO = "BAR"
}
```

or keep `user` and `environment` in `config`, but only allow those keys to be updated:

```hcl
config {
  limits.cpu = 2
  user.role = "mysql"
  environment.FOO = "BAR"
}
```

I think the above would wreck havoc with the Terraform state...

* What about exporting (computed) configuration? For example:

```hcl
resource "lxd_profile" "foo" {
  config {
    environment.FOO = "BAR"
  }
}

resource "lxd_container" "foo" {
  profiles = ["default", "foo"]
}
```

Should `lxd_container.foo` have access to `environment.FOO`? This might be useful for provisioning.

Merging exported/computed configuration with user-specified configuration might be a bear. Perhaps a better option is to separate input and output. For example, user-defined configuration goes in `config`, but all configuration is available in a read-only exported `container_config` block.

jtopjian:
Revisiting this after a week or so, I'm leaning on the following:

* Allow updates to anything LXD wants to update. Discretion is left to the user.
  
  This is the easiest way to handle things since anything in `config` would just be updated. "User defined" keys would easily be handled here, too.

[...]

PR #51, opened 2017-06-06
[WIP] Experiment with updatable config

sl1pm4t:
Testing out Updatable config values. 

Currently limited to the `limits.` namespace that has been spun out to it's own resource attribute.

Relates to #21

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -83,6 +83,12 @@ func resourceLxdContainer() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"limits": &schema.Schema{
+				Type:     schema.TypeMap,
+				Optional: true,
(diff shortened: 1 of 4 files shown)


lxd/resource_lxd_container.go
```diff
@@ -174,6 +180,8 @@ func resourceLxdContainerCreate(d *schema.ResourceData, meta interface{}) error
 		image = imgParts[1]
 	}
 	config := resourceLxdConfigMap(d.Get("config"))
+	config = resourceLxdConfigMapAppend(d.Get("limits"), config, "limits.")
```
jtopjian: So between the above two lines of code, if someone specified `limits.cpu` in the `config` block and then specified `cpu` in the `limits` block, the latter should overwrite the former? If so 👍

PR #56, opened 2017-06-06
Utilize lxdhelper library for certain functions

jtopjian:


jtopjian:
@sl1pm4t Looks like the `LXD_PASSWORD` environment variable isn't being set?

sl1pm4t:
@jtopjian  Strange about LXD_PASSWORD... that part worked OK in my last build (an hour ago)

jtopjian:
Here are the Travis settings from my point of view. I wonder if it's a permissions issue:

screen shot 2017-06-06 at 1 30 39 pm

sl1pm4t:
Must be this travis change:
https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions

I guess there's no reason to hide the password, seeing as the env is ephemeral.

Issue #53, opened 2017-06-06
high CPU consumption when using file block


shantanugadgil:
Terraform version: Terraform v0.9.6
terraform-provider-lxd: 0.9.5 beta1

I tried to deploy a binary file (~ 17MB) using the following **file** block inside an **lxd_container**
```
file {
  content     = "${file("config/minio.bin")}"
  target_file = "/usr/sbin/minio"
  uid         = 0
  gid         = 0
  mode        = "0755"
}
```

When I do "terraform plan" the terraform process can be seen using all the CPU; ~150% CPU on my two core machine.

This stays at high CPU for quite a few minutes and eventually freezes the machine.

My current workaround is to use the internal "file" provisioner which copies the file onto the lxd server, and then a "remote-exec" to do an "lxc file push ..." to copy the file into the LXD.

Thought: Can the file contents shown on the screen be reduced to, say a smaller length, say 1024 bytes? rather than printing out the entire file upon change?

Thanks and Regards,
Shantanu

PR #63, opened 2017-06-10
Hash the contents of files

jtopjian:
This commit hashes the contents of a file before it stores it
in state. This is useful for large files.

For #53

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -1,6 +1,8 @@
 package lxd
 
 import (
+	"crypto/sha1"
+	"encoding/hex"
 	"fmt"
 	"log"
 	"os"
@@ -105,6 +107,15 @@ func resourceLxdContainer() *schema.Resource {
 						"content": &schema.Schema{
 							Type:     schema.TypeString,
 							Required: true,
+							StateFunc: func(v interface{}) string {
+								switch v.(type) {
+								case string:
+									hash := sha1.Sum([]byte(v.(string)))
+									return hex.EncodeToString(hash[:])
+								default:
+									return ""
+								}
+							},
 						},
 
 						"target_file": &schema.Schema{


jtopjian:
@shantanugadgil This should solve #53 for you. Let me know if you're able to test it and if it resolves the problem you're seeing.

shantanugadgil:
Terraform: Terraform v0.9.8

I tried but doesn't work, it still prints out the long string (17 MB long) and the same happens; high CPU during plan.
I went ahead and did apply anyway; it failed with: (s3 is the name of my minio `lxd_container`)
```
* lxd_container.s3: 1 error(s) occurred:

* lxd_container.s3: reading body session shutdown
```

This is how I checked out and built the branch:
```
git clone https://github.com/jtopjian/terraform-provider-lxd.git
cd terraform-provider-lxd/
git checkout container-file-hash
make
make build dev
# ls terraform-provider-lxd
terraform-provider-lxd
```

sl1pm4t:
HI @shantanugadgil - did you copy the newly built binary into your `$PATH`? 
Perhaps try `cd $GOPATH/src/github.com/jtopjian/terraform-provider-lxd; go install -v`

I tried out this fix and it worked for me - before the fix a 12Mb file would hang terraform during a plan and afterwards the plan completed in a short time. 

Also, after `terraform apply` the file integrity was verified with an `md5sum` of the file inside vs outside the container.

shantanugadgil:
@sl1pm4t  my build machine and tf submit hosts are different.
So I scp the `terraform-provider-lxd` and place it in the same directory as by `terraform` binary.

sl1pm4t:
Yes, this seems to only be a partial fix. I was able to reproduce further issues here.
Terraform still hangs when running `plan` on an existing resource where a large file was pushed.

shantanugadgil:
Yes, another side effect of this problem is that the file `terraform.tfstate` has gone to ~52MB.
Initially it would be < 1MB

jtopjian:
This is looking like a bug in Terraform core. The changed value is being displayed during run, but not stored in the state. This seems to be only happening for nested elements of a List/Set.

This is applying correctly whenever applied to a top-level schema item. For example,

```go
"name": &schema.Schema{
  Type:     schema.TypeString,
  ForceNew: true,
  Required: true,
  StateFunc: func(v interface{}) string {
    return v.(string) + "!!!"
  },
},
```

I'll inquire about this on Monday, but if it is a core bug there are two things we can do:

1. Leave this as-is
2. Modify the `file.content` argument to do a validation check. Anything over a certain size will trigger an error and not proceed.

jtopjian:
Another possibility is to create a first-class `lxd_file` resource.

PR #64, opened 2017-06-10
Fix hashing of file contents in Terraform state

jtopjian:
This should work around the issue discovered in #63

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -115,13 +115,8 @@ func resourceLxdContainer() *schema.Resource {
 							Type:     schema.TypeString,
 							Required: true,
 							StateFunc: func(v interface{}) string {
-								switch v.(type) {
-								case string:
-									hash := sha1.Sum([]byte(v.(string)))
-									return hex.EncodeToString(hash[:])
-								default:
-									return ""
-								}
+								hash := sha1.Sum([]byte(v.(string)))
+								return hex.EncodeToString(hash[:])
 							},
 						},
 
@@ -242,13 +237,19 @@ func resourceLxdContainerCreate(d *schema.ResourceData, meta interface{}) error 
 
 	d.SetId(name)
 
-	// Upload any files, if specified
-	if v, ok := d.GetOk("file"); ok {
-		for _, v := range v.([]interface{}) {
-			if err := resourceLxdContainerUploadFile(client, name, v); err != nil {
+	// Upload any files, if specified,
+	// and set the contents to a hash in the State
+	if files, ok := d.GetOk("file"); ok {
+		for _, v := range files.([]interface{}) {
+			file := v.(map[string]interface{})
+			if err := resourceLxdContainerUploadFile(client, name, file); err != nil {
 				return err
 			}
+			hash := sha1.Sum([]byte(file["content"].(string)))
+			file["content"] = hex.EncodeToString(hash[:])
 		}
+
+		d.Set("file", files)
 	}
 
 	return resourceLxdContainerRead(d, meta)


jtopjian:
cc @shantanugadgil

jtopjian:
I'm still going to investigate an `lxd_file` resource. This is just to workaround the bug of #63

shantanugadgil:
@jtopjian I'll give this a try and let you know.
+1 for `lxd_file` if you do decide on it.

shantanugadgil:
@jtopjian  I am no longer sure why the same symptom keeps occurring on my setup.

I used the new branch as can be seen here `jtopjian:container-file-hash-fix`

I think I'll try to complete other things before I get back to experimenting with large files in the `file` block.

jtopjian:
@shantanugadgil which symptoms? Is the Terraform state file still large?

shantanugadgil:
Yes, `terraform.tfstate` file grows large (~52 MB).
I think it is filled with the "\x000" byte values of the file itself.
And quite often the `apply` fails with a "connection lost" message.

jtopjian:
Yeah, that's definitely strange. With this patch applied, I see the following in my state file:

```json
"lxd_container.test1": {
    "type": "lxd_container",
    "depends_on": [],
    "primary": {
        "id": "test1",
        "attributes": {
            "config.%": "1",
            "config.limits.memory": "2GB",
            "ephemeral": "false",
            "file.#": "1",
            "file.0.content": "8c206a1a87599f532ce68675536f0b1546900d7a",
            "file.0.create_directories": "false",
            "file.0.gid": "0",
            "file.0.mode": "0644",
            "file.0.target_file": "/tmp/foo.txt",
            "file.0.uid": "0",
            "id": "test1",
            "image": "ubuntu",
            "ip_address": "10.3.184.70",
            "mac_address": "00:16:3e:86:cc:95",
            "name": "test1",
            "privileged": "false",
            "profiles.#": "1",
            "profiles.0": "default",
            "status": "Running"
        },
        "meta": {},
        "tainted": false
    },
    "deposed": [],
    "provider": ""
}```

**Comment by shantanugadgil** [comment]
FWIW, this my example lxd_container definiton

```hcl
resource "lxd_container" "s3" {
  remote    = "..."
  name      = "..."
  image     = "..."
  ephemeral = false
  profiles  = ["${var.profiles}"]

    file {
      content     = "${file("config/minio.bin")}"
      target_file = "/tmp/minio"
      uid         = 0
      gid         = 0
      mode        = "0755"
    }
}
```

The minio binary is from here:
https://dl.minio.io/server/minio/release/linux-amd64/minio

sl1pm4t:
FYI @jtopjian  - as another workaround for this, I experimented with enabling the `Sensitive` flag on the `file` `contents` attribute.
With that change, the Terraform no longer burns CPU cycles and dumps the contents of the file to output, it does however still store the full contents of the file in the .tfstate file, so still not ideal.

shantanugadgil:
cross-referencing #53 for high CPU usage

jtopjian:
@sl1pm4t Are you still seeing high CPU usage with this change? And just to confirm: the hash is being stored in the state file for you?

Does adding `Sensitive` resolve the CPU issue for you?

sl1pm4t:
@jtopjian - I think was only seeing high CPU when the file contents were being dumped to output, and so the `sensitive` flag fixed that - however I'll double check when I get a chance later today.

shantanugadgil:
@sl1pm4t Did you get a chance to experiment again?

Thanks,
Shantanu

sl1pm4t:
Sorry everyone - life got busy for a while.
Looking at this now.

sl1pm4t:
With the sensitive fix I don't see high CPU during `apply` or `plan` phases.

sl1pm4t:
Actually - I also tested without the `sensitive` flag, but built against `Terraform v0.10.0-dev (f40df21988572d5dfa4830bb5685a0c3134271ba)` build and also didn't reproduce high CPU. 

So maybe it's not needed now, because the issue @jtopjian identified is fixed in Terraform.

jtopjian:
> Sorry everyone - life got busy for a while.

No worries at all :smile:

PR #65, opened 2017-06-11
Allow update of container limits

sl1pm4t:
This allows updating the resource limits of a container. 
This PR replace #51 

A new container resource attribute `limits` is added, that behaves similar to the `config` attribute, but allows value changes to be updated without a container destroy + recreate cycle.

**Example config:**
```hcl
resource "lxd_container" "c1" {
  name = "c1"
  image = "ubuntu"

  limits {
    "cpu" = "2"
  }
}
```

### DIFF
--- a/lxd/shared.go
+++ b/lxd/shared.go
@@ -64,19 +64,24 @@ func resourceLxdConfigMap(c interface{}) map[string]string {
 	return config
 }
 
-func resourceLxdConfigMapAppend(c interface{}, config map[string]string, namespace string) map[string]string {
+// resourceLxdConfigMapAppend appends a map of configuration values
+// to an existing map. All appended config values are prefixed
+// with the config namespace.
+func resourceLxdConfigMapAppend(config map[string]string, append interface{}, namespace string) map[string]string {
 	if config == nil {
-		config = make(map[string]string)
+		panic("config is nil")
 	}
 
 	if string(namespace[len(namespace)-1]) != "." {
 		namespace += "."
 	}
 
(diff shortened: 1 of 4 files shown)


shantanugadgil:
Just out-of-curiosity:

Moving things from free-form `config` section into a specific `limits` is fine for **_now_**, but what about the long-term, where other config parameters need to be updated, which **don't** really need a re-create?
Example: `boot.autostart`

* I did notice that you have mentioned that config values like `autostart` will cause a re-create for now, but I hope that's just a temporary thing, and over time everything (nearly) that be done from the `lxc` command line can be done via `terraform-provider-lxd`

Regards,
Shantanu

lxd/shared.go
```diff
@@ -64,6 +64,26 @@ func resourceLxdConfigMap(c interface{}) map[string]string {
 	return config
 }
 
+func resourceLxdConfigMapAppend(c interface{}, config map[string]string, namespace string) map[string]string {
```
jtopjian: Possible nit-picking, but would this be better if it was `resourceLxdConfigMapAppend(config, c, namespace)`?

`config` is the map being appended to, `c` is the smaller set with new data.

jtopjian:
@shantanugadgil The way I view this is just the first implementation of update-able configuration. It's certainly possible to continue adding this pattern when one has the time :)

lxd/shared.go
sl1pm4t: Thanks @jtopjian - I like your logic.

Issue #67, opened 2017-06-24
changes to file order


jfgibbins:
As a suggestion, place lines 202-209 of resource_lxd_container.go to line 180 so that files are copied before container starts.  This will allow, among other things, placing files, such as rt_tables, and if-up.d files in place before networking kicks in.  Already changed my local copy.  If I was more familiar with git, I'd push/pull whatever for you to merge in.

shantanugadgil:
if the change is small enough, you can just make online (in the webbrowser); github's online file editor will lead you through the steps. 😄 

otherwise:
fork the repo into your github account (so you have write access (push right) to it)
pull from your account onto your disk
make local changes; commit files (s), then push the file(s) back to your on github account
create pull request from your github account to the main account.

HTH 👍

jtopjian:
Fixed by #76

PR #68, opened 2017-06-24
per #67

jfgibbins:
executes file push before starting container, so that files are available for startup routines

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -212,6 +212,15 @@ func resourceLxdContainerCreate(d *schema.ResourceData, meta interface{}) error 
 		return err
 	}
 
+	// Upload any files, if specified
+	if v, ok := d.GetOk("file"); ok {
+		for _, v := range v.([]interface{}) {
+			if err := resourceLxdContainerUploadFile(client, name, v); err != nil {
+				return err
+			}
+		}
+	}
+
 	// Start container
 	_, err = client.Action(name, shared.Start, -1, false, false)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


sl1pm4t:
Hi @jfgibbins - thanks for your PR, although it looks like it had unintended side-effects and has broken the tests. Do you know if this use case (uploading files before container start) works outside of Terraform?

jfgibbins:
Yes it does.  lxc push/pull works regardless of whether the container is working.  Also, fyi, I had already made the change locally, and it worked like a charm.  I use it to push files into /etc/network/if-up.d and to /etc/iproute2/rt_tables so that there available before the initial network start.  Please double check my edit, but it should just have been moving the file upload stanza from after container startup, to be between creation and startup.

jfgibbins:
this is what the section looks like in the local copy that was compiled.  

        // Wait for the LXC container to be created
        err = client.WaitForSuccess(resp.Operation)
        if err != nil {
                return err
        }

        // Upload any files, if specified
        if v, ok := d.GetOk("file"); ok {
                for _, v := range v.([]interface{}) {
                        if err := resourceLxdContainerUploadFile(client, name, v); err != nil {
                                return err
                        }
                }
        }

        // Start container
        _, err = client.Action(name, shared.Start, -1, false, false)
        if err != nil {
                // Container has been created, but daemon rejected start request
                return err
        }

        // Wait until the container is in a Running state
        stateConf := &resource.StateChangeConf{
                Target:     []string{"Running"},
                Refresh:    resourceLxdContainerRefresh(client, name),
                Timeout:    3 * time.Minute,
                Delay:      refreshInterval,
                MinTimeout: 3 * time.Second,
        }

        if _, err = stateConf.WaitForState(); err != nil {
                return fmt.Errorf("Error waiting for container (%s) to become active: %s", name, err)
        }

        d.SetId(name)

        return resourceLxdContainerRead(d, meta)

jtopjian:
Superseded by #76

jtopjian:
@jfgibbins Thanks for submitting this. There were some changes to this block of code between when you originally submitted this and now, so I went ahead and merged an updated version. I also fixed the unit tests so everything passed in testing.

jfgibbins:
Very cool.  Glad to know also why the tests failed.  So, sort of makes my first PR ever, lol.

PR #72, opened 2017-07-16
Adding terraform files to build test environments

jtopjian:


jtopjian:
Looks like the test failures have been fixed in the new client. I might also be updating the `deploy.sh` script depending on the outcome of the new client errors... this PR can wait until that's all sorted out.

PR #73, opened 2017-07-16
Sleep during container transition

jtopjian:


jtopjian:
Nope. That didn't work :/

PR #76, opened 2017-07-19
Upload Files Before Container Starts

jtopjian:
This commit uploads files before the container starts, which is
useful for bootstrapping the container with configuration files
and startup scripts.

Supersedes #68

### DIFF
--- a/lxd/resource_lxd_container_test.go
+++ b/lxd/resource_lxd_container_test.go
@@ -660,7 +660,7 @@ resource "lxd_container" "container1" {
 
   file {
     content = "Hello, World!\n"
-    target_file = "/tmp/foo/bar.txt"
+    target_file = "/foo/bar.txt"
     mode = "0644"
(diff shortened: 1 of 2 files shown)


lxd/resource_lxd_container_test.go
```diff
@@ -660,7 +660,7 @@ resource "lxd_container" "container1" {
 
   file {
     content = "Hello, World!\n"
-    target_file = "/tmp/foo/bar.txt"
+    target_file = "/foo/bar.txt"
```
sl1pm4t: was this the cause of the test failures before? 
does `/tmp` not exist before container start?

jtopjian: Yep :)

sl1pm4t: Nice one, seems obvious now but I didn't think to check that.

Thanks!

Issue #66, opened 2017-06-11
[feature request] ability to create already deleted file


shantanugadgil:
During my experiments, I forgot to tune down `parallelism` to a good working value, and ended up with a mixed-up state.

The file "data.cfg" never got created the first time round.

To make TF recreate the file, I changed the order of the file blocks inside the `lxd_container`.
It detects the change alright, but errors out with the following

```
Error applying plan:
1 error(s) occurred:
* lxd_container.test1: 1 error(s) occurred:
* lxd_container.test1: Could not delete file /etc/file/path/data.cfg: not found
```

Would there be any way to taint the `file` block inside the resource (I am not sure there is a way).

OR

Can there be an extra option inside the `file` block (or `lxd_container`) to force re-create files which are not found in the target container.

Regards,
Shantanu

jtopjian:
Both options might be possible. I'll look into them this week.

jtopjian:
Just a quick update: I'm going to wait until #64 is resolved before working on this to prevent too many modifications in one area at the same time.

shantanugadgil:
Thanks @jtopjian 
@sl1pm4t  Any chances of a new release build from master after merging the few PRs still not in master?

Thanks and Regards,
Shantanu

jtopjian:
This should be fixed in #77 

If an error occurs during creation before the file was uploaded, the file should be uploaded upon doing another apply.

PR #77, opened 2017-07-19
Use Partial for Container Creation

jtopjian:
This commit uses "partial" for container creation. This should enable
containers to resume uploading files if there was previously an error
during creation, without having to do a full recreate of the container.

For #66


jtopjian:
I'll rebase/redo this once #76 is merged. Just wanted to get the research/draft done and see if Travis is happy with the change.

PR #86, opened 2017-09-09
Updating Travis to use Go 1.9

### LINKED ISSUE #85: libacl and Go 1.9 dependencies
A recent change to upstream LXD now requires that `libacl1-dev` is installed in order for `terraform-provider-lxd` to build. 

I've opened an issue at https://github.com/lxc/lxd/issues/3787 to see if libacl is in fact required for applications that consume LXD. If it is required, we'll have to update Travis and the build documentation accordingly.

As well, Terraform core now requires Go 1.9 to build.

jtopjian:
Fixes #85

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_install:
 - lxc image list
 
 go:
-- 1.8
+- 1.9
 
 matrix:
   fast_finish: true


jtopjian:
hrm... looks like there's some other package updates required. I'm going to close this one and submit a full PR.

PR #90, opened 2017-09-10
Fix test and travis build

mjrider:
- switch to go 1.9, terraform requires this 
- fixed the provider_test
- yamllint'ed .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,34 @@
+---
 language: go
 sudo: true
 dist: trusty
 
 env:
-- TF_ACC=1
+  - TF_ACC=1 \
+   LXD_REMOTE=travis \
+   LXD_ADDR=localhost \
+   LXD_PORT=8443 \
+   LXD_GENERATE_CLIENT_CERTS=true \
+   LXD_ACCEPT_SERVER_CERTIFICATE=true \
+   LXD_SCHEME=https \
+   LXD_PASSWORD=the-password \
 
 before_install:
-- sudo add-apt-repository -y ppa:ubuntu-lxc/lxd-stable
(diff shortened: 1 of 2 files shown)


jtopjian:
@mjrider Thanks for this, but please see #87 :)

The `SaveConfig` piece looks like it's a better solution to what I have, though. I was wondering about that part and felt wrong modifying the test case to account for it.

jtopjian:
The travis settings look like they're a better fit, too. @sl1pm4t has them configured within Travis, but having them in the `.travis.yml` file makes them more accessible.

I'm in favor of closing #87 for this - I'll let @sl1pm4t decide that one, though.

mjrider:
@jtopjian yeah, sorry, i was working on fixing the tests for a few days, and forget to check if somebody else also did same work in the meantime.

i tried to run the tests it complained that RemoteConfig and SaveConfig were not defined, and it lookt to me that the api of lxd changed. So i changed the test case to compensate for that.

Maybe it would be a good improvement to fix the vendored code with govendor, other terraform providers seem to use that.

jtopjian:
> @jtopjian yeah, sorry, i was working on fixing the tests for a few days, and forget to check if somebody else also did same work in the meantime.

No problem at all. After review, I think this PR is much better. I would merge it, but just want @sl1pm4t to sign off on the travis changes.

> Maybe it would be a good improvement to fix the vendored code with govendor, other terraform providers seem to use that.

We've opted to not version so it forces us to keep up with LXD development since it changes so frequently. Versioning would set us back quite a ways.

sl1pm4t:
Thanks @mjrider + @jtopjian

PR #91, opened 2017-09-11
Add profile import

mjrider:
This adds the option to import lxd_profiles.
it also fixes a bug in the resourceLxdProfileRead function, which didn't read the devices from lxd because of a error in the data formating.

I could use some help with creating test cases for this, that is why i marked it with WIP

### DIFF
--- a/lxd/import_resource_lxd_profile_test.go
+++ b/lxd/import_resource_lxd_profile_test.go
@@ -27,9 +27,3 @@ func TestLXDProfile_importBasic(t *testing.T) {
 		},
 	})
 }
-
-const testLXDProfileDefault = `
-resource "lxd_profile" "default" {
-	name = "default"
-}
-`
--- a/lxd/import_resource_lxd_profile_test.go
+++ b/lxd/import_resource_lxd_profile_test.go
@@ -1,0 +1,35 @@
+package lxd
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/dustinkirkland/golang-petname"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestLXDProfile_importBasic(t *testing.T) {
+	profileName := strings.ToLower(petname.Generate(2, "-"))
+	resourceName := "lxd_profile.profile1"
+
+	resource.Test(t, resource.TestCase{
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccProfile_basic(profileName),
+			},
+
+			resource.TestStep{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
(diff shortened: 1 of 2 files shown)


jtopjian:
@mjrider Generally `import` tests are done in a separate `import_*.go` file, such as `import_lxd_profile_test.go`.

You can see a simple example here:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/import_openstack_compute_flavor_v2_test.go

It's best if the import test re-uses the "basic" acceptance test of a profile.

Let me know if that helps :)

mjrider:
added a basic testcase, it passes go test locally, but for travis to test it we first need #90 to be merged

mjrider:
Found a issue when importing resources from a non-default remote, i'll have to look into that

lxd/import_resource_lxd_profile_test.go
```diff
@@ -0,0 +1,35 @@
...
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+const testLXDProfileDefault = `
+resource "lxd_profile" "default" {
+	name = "default"
+}
```
jtopjian: This can safely be removed.

lxd/resource_lxd_profile.go
```diff
@@ -209,3 +227,30 @@ func resourceLxdProfileExists(d *schema.ResourceData, meta interface{}) (exists
...
+func resourceLxdProfileImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	p := meta.(*LxdProvider)
+
+	ids := strings.SplitN(d.Id(), ":", 2)
+	name := ids[0]
+	d.SetId(name)
+
+	if len(ids) == 2 {
+		d.Set("remote", ids[1])
+	}
```
jtopjian: Is it correct to say that in order to import the profile, you must do:

```shell
$ terraform import lxd_profile.profile_name profile_name:remote_name
```

Would it make more sense to reverse the import name and do:

```shell
$ terraform import lxd_profile.profile_name remote_name:profile_name
```

?

I think you can use [`conf.ParseRemote`](https://github.com/lxc/lxd/blob/daff16273d9ff353ed30bbecd9d368555111c341/lxc/config/remote.go#L20) to easily split the name.

lxd/import_resource_lxd_profile_test.go
mjrider: i agree, left over from refactoring..

lxd/resource_lxd_profile.go
mjrider: mmmm maybe in this scope it makes sense to have the optional argument as prefix, because the remote  is optional, i'll make a change for this later this week

shantanugadgil: [remote_name:]profile_name makes more sense to me too! 👍

jtopjian:
Looks good to me. Thanks for adding this!

Issue #93, opened 2017-09-12
ip_address returned


jfgibbins:
how is the lxd provider determining the ip address of the container to use for connections as well as for saving in the tfstate file?  Most of my containers are multihomed, and inevitably, the provider always chooses the wrong one.  While I can and do set the connection for remote-exec, that information can't be carried over to other resources.

jfgibbins:
Manually setting the ip_address, would be a trade off with having to set connection for remote-exec, so even money, but ideally combined with #21, one of the connections for network-config could be set as "default" for the purposes of the resource.

jfgibbins:
@jtopjian last detected would definitely be a problem on any container doing any sort of containers, like docker or lxd inside a container, though make sure and leave the tag open, for example some may use, eth1, other's like mine, use trunks, so it's actually eth0.1000, at least until I can get cloud-init to do naming on trunks and get it named mgmt...so you'll have that coded and out in the next hour?  ;)

jtopjian:
Right, the idea would be to configure a container like so:

```hcl
resource "lxd_container" "foo" {
  config {
    user.access_network = "eth0.1000"
  }
}
```

Inside the `for` loop I linked to earlier, there would be an extra piece of logic to see if `user.access_network` was defined and if so, set `ip_address` and `mac_address` to what was defined on `eth0.1000`.

The benefit of using `user` is that LXD reserves it for miscellaneous metadata. As well, keeping that data set in the LXD container's metadata makes it agnostic to cloud-init and such, so it would work with any distribution (with or without cloud-init).

The name `access_network` is taken from the same type of implementation I did in the Terraform OpenStack provider. Perhaps for LXD, `access_interface` might be more applicable.

> so you'll have that coded and out in the next hour?

Not quite :) but it should be a pretty easy thing to put in place. I'll try to get to it within the next week.

jfgibbins:
Sounds like just what I need.  As it stands, my docker containers, end up with 4 addresses, public, mgmt, docker0 and docker_gwbridge with some custom source-routing thrown in.  Hopefully I don't ever decide to run docker on a dmz services container, cause then it would be 5 with the private_ip as well.  Yes, I'm a network geek.

jfgibbins:
@jtopjian Just saw that they released an LXD builder for packer a couple days ago.  Going to play with it and test it for a few days while this update is pending.  Initial tests show some promise.

PR #94, opened 2017-09-16
Container Access Interface

jtopjian:
This commit enables a user to specify which interface will
be used for the ip_address and mac_address in a container
by specifying a user.access_interface config setting.

I opted to use a config setting rather than a first-class Terraform parameter since `user.access_interface` should be able to be set in a profile and then be applied to all containers using that profile.

For #93

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -342,8 +342,8 @@ func resourceLxdContainerRead(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return err
 	}
+	log.Printf("[DEBUG] Retrieved container state %s:\n%#v", name, state)
 
-	log.Printf("[DEBUG] Retrieved container config %s:\n%#v", name, container.Config)
 	for k, v := range container.Config {
 		if strings.Contains(k, "limits.") {
 			log.Printf("[DEBUG] Setting limit %s: %s", k, v)
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -354,13 +354,33 @@ func resourceLxdContainerRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("status", container.Status)
 
 	sshIP := ""
-	for iface, net := range state.Network {
-		if iface != "lo" {
-			for _, ip := range net.Addresses {
-				if ip.Family == "inet" {
-					d.Set("ip_address", ip.Address)
-					sshIP = ip.Address
-					d.Set("mac_address", net.Hwaddr)
+	// First see if there was an access_interface set.
+	// If there was, base ip_address and mac_address off of it.
+	var aiFound bool
+	if ai, ok := container.Config["user.access_interface"]; ok {
+		net := state.Network[ai]
+		log.Printf("[DEBUG] FOO: %#v", net)
+		for _, ip := range net.Addresses {
+			if ip.Family == "inet" {
+				aiFound = true
+				d.Set("ip_address", ip.Address)
+				sshIP = ip.Address
+				d.Set("mac_address", net.Hwaddr)
+			}
+		}
+	}
+
+	// If the above wasn't successful, try to automatically
+	// determine the ip_address and mac_address.
+	if !aiFound {
+		for iface, net := range state.Network {
+			if iface != "lo" {
+				for _, ip := range net.Addresses {
+					if ip.Family == "inet" {
+						d.Set("ip_address", ip.Address)
+						sshIP = ip.Address
+						d.Set("mac_address", net.Hwaddr)
+					}
 				}
 			}
 		}
--- a/lxd/resource_lxd_container_test.go
+++ b/lxd/resource_lxd_container_test.go
@@ -325,6 +325,27 @@ func TestAccContainer_configLimits(t *testing.T) {
 	})
 }
 
+func TestAccContainer_accessInterface(t *testing.T) {
+	var container api.Container
+	networkName := strings.ToLower(petname.Generate(2, "-"))
+	containerName := strings.ToLower(petname.Generate(2, "-"))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccContainer_accessInterface(networkName, containerName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccContainerRunning(t, "lxd_container.container1", &container),
+					resource.TestCheckResourceAttr("lxd_container.container1", "name", containerName),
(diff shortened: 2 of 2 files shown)


jfgibbins:
@jtopjian Looks like you've been busy.  Funny when the test code is longer than the code being tested.

jfgibbins:
@jtopjian I'm guessing still some work to do, but I compiled your branch to start some testing.  It's not always successful in grabbing the correct information, falling back to the old method.  In addition, destruction of containers no longer works.  It appears to be hanging itself at the step where the network port on open-vswitch is being deleted.

jtopjian:
@jfgibbins If it's falling back to the old method, that means it wasn't able to find an IP address on the NIC. Sometimes this is because the Linux distribution only sets DHCP on the first interface by default, so other interfaces won't automatically come up on boot. You might have to create a custom LXD image to support DHCP on multiple NICs.

If DHCP isn't the case, can you post the debug logs to a github gist (gist.github.com). You can enable debug logging with:

```
$ TF_LOGS=DEBUG terraform apply
```

jfgibbins:
@jtopjian I'm not using dhcp.  everything is statically assigned using cloud-init.  Give me a few and let me spin it up with some debugs for you.

jfgibbins:
@jtopjian 
[debug.txt](https://github.com/sl1pm4t/terraform-provider-lxd/files/1308544/debug.txt)
Here is the one for creating.  Waiting on the destroy attempt to timeout for those logs.  Forthcoming

jfgibbins:
@jtopjian 
[debug2.txt](https://github.com/sl1pm4t/terraform-provider-lxd/files/1308546/debug2.txt)
And this is for the destroy attempt.

jtopjian:
@jfgibbins Thanks for the debug logs. I see that we're not actually printing any debug information for the network, so I've just pushed a commit to fix this.

If you can, can you provide debug logs again - I'm really sorry about that.

> everything is statically assigned using cloud-init

I have a feeling this is where the problem is. cloud-init isn't part of LXD or a linux distribution. It's a helpful userspace program that will do post-boot configurations. It can consume various metadata that different systems can publish (LXD, OpenStack, AWS).

cloud-init is setting the network information well after the container has booted and well after the `Read` function in Terraform.

I have a feeling that if you run `terraform refresh` after the container has been created and one or two minutes later to let cloud-init do its thing, `ip_address` might show the right information.

jfgibbins:
It might be a few, need to get dinner on before Jessie heads to work.  She's got back to back night shifts at the hospital. :(

jtopjian:
No problem at all - there's no rush.

jfgibbins:
@jtopjian Can you double check me, what's the correct command for compiling the right branch, or did you merge it into your master?

jtopjian:
It's not merged into master yet. To compile this branch, you'll want to do:

```shell
$ go get github.com/sl1pm4t/terraform-provider-lxd
$ cd ~/go/src/github.com/sl1pm4t/terraform-provider-lxd
$ git remote add jtopjian https://github.com/jtopjian/terraform-provider-lxd
$ git fetch jtopjian
$ git checkout jtopjian/container-access-interface
$ make build
```

That's off the top of my head, so let me know if there's a typo. We should add that to the README.

jfgibbins:
FYI, I'm wondering if I compiled the right version the first time around.  If I didn't, then the source of the destroy problem is in sl1p's master.  The problem did not exist in v0.10.0-beta2, which is what I was using this morning.  I'm going to grab 10.4 that you built while you look at the logs.

[debug.txt](https://github.com/sl1pm4t/terraform-provider-lxd/files/1308582/debug.txt) - apply
[debug2.txt](https://github.com/sl1pm4t/terraform-provider-lxd/files/1308587/debug2.txt) - destroy

Here are the debugs, as well as what ovs-vsctl reports during the destroy.  Version 2.6.1 of ovs
        Port "dock-work01"
            Interface "dock-work01"
                error: "could not open network device dock-work01 (No such device)"
        Port "dock-sup02"
            Interface "dock-sup02"
                error: "could not open network device dock-sup02 (No such device)"
        Port "dock-sup01"
            Interface "dock-sup01"
                error: "could not open network device dock-sup01 (No such device)"
        Port "dock-work02"
            Interface "dock-work02"
                error: "could not open network device dock-work02 (No such device)"

jtopjian:
If this wasn't happening in the last release, it could be a number of things. I'm leaning on an LXD client update. Are you able to reproduce this problem by manually launching and destroying containers?

jfgibbins:
manual launches and deletes are successful.  I'm testing with 10.4 right now, to see if if was a change between 10.0.0 beta and 10.4

jtopjian:
When you doing the manual launches, you're including all of the network-data and other configurations that would be included in Terraform?

jfgibbins:
I'm going to go through, check versions, update, etc, and narrow down, cause 10.4 crashed lxd, but then I noticed terraform itself was just updated.  Very possible the problem, I'll let you know.

jfgibbins:
@jtopjian Ok, I've tested.  Might add download new version of terraform to the docs when updating client.  I was on 10.2, upgraded to 10.5, used the latest compiled version of provider that you helped me with.  Seems to be working.  Remember, I'm an idiot. :)  Do love the up update.  Lots of terraform code and packer builds(btw, they finally released lxd builder for packer) to work on now.  Thanks, and I promise, I'll eventually learn go.

jfgibbins:
@jtopjian your update is working well.  My fault for not updating terraform to the latest first.  I should have known better.  Appreciate the help.

jtopjian:
@jfgibbins It's no problem at all - I wouldn't have thought to update to the latest Terraform core, so it was a good idea to try.

So is this working with your cloud-init settings and not having to do a `terraform refresh`? If so, that's great news, too.

I'll let @sl1pm4t give this a review and merge when he's ready.

jfgibbins:
so far it seems to be grabbing the right address without any timing problems.  I'll scale it up to a larger count to push the system, see if lag causes any issues, but go ahead and assume its good for now, and I'll let you know otherwise.  Makes me wonder how many concurrent I can push at once before it slows down my system.  32 cores Haswell-E  :D

jfgibbins:
@jtopjian Just pushed a terraform apply, with 4 different types of container resources, totaling 60 instances, images pre-built with packer.  All ip_address' came out correct.  BTW, my cpu usage went from 1.5% to 2.5%, probably bound by drives.  Really need to drop in about 8 SSDs to add to the ZFS, little for L2ARC, a couple write oriented ones for ZIL.  You know, toys.

jfgibbins:
@jtopjian just an fyi, doesn't really matter for me, but adding access_interface to existing , does of course cause the container to be destroyed/recreated.  Not an issue I don't think, and  I'm sure expected as it's under config.

jtopjian:
Bah. Looks like either Travis is having issues or an upstream change to LXD triggered something.

jtopjian:
Tests are working fine on a clean 16.04 environment, though the environment is using the btrfs backend and not the directory backend. I can dig into this some more when I get a moment.

PR #95, opened 2017-09-25
Fix profile importing with older versions of LXD

mjrider:
On Ubuntu xenial with lxd version 2.0.9 importing a profile resulted in de following device 
```
  device.# = 1
  device.0.name = eth0
  device.0.properties.% = 3
  device.0.properties.name = eth0
  device.0.properties.nictype = bridged
  device.0.properties.parent = br2
  device.0.type = nic
```

where  **device.0.properties.name** is obvious 'wrong' 

added 2 testcases, one for config and one for devices.
on recent lxd releases, like i'm running on my laptop, this isn't an issue but it is in my production environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
   - sudo apt-get install -y snapd
   - sudo snap install lxd
   - export PATH=/snap/bin/:${PATH}
-  - sleep 10 # give lxd time to start
+  - sudo lxd waitready --timeout 60
   - sudo lxd init --auto --trust-password="${LXD_PASSWORD}" --network-port=8443 --network-address='[::]'
   - sudo chmod 777 /var/snap/lxd/common/lxd/unix.socket
   - 'lxc image copy ubuntu:t local: --alias=ubuntu'
--- a/lxd/import_resource_lxd_profile_test.go
+++ b/lxd/import_resource_lxd_profile_test.go
@@ -27,3 +27,43 @@ func TestLXDProfile_importBasic(t *testing.T) {
 		},
 	})
 }
+
+func TestLXDProfile_importConfig(t *testing.T) {
+	profileName := strings.ToLower(petname.Generate(2, "-"))
+	resourceName := "lxd_profile.profile1"
+
+	resource.Test(t, resource.TestCase{
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccProfile_config(profileName),
+			},
+
+			resource.TestStep{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestLXDProfile_importDevice(t *testing.T) {
+	profileName := strings.ToLower(petname.Generate(2, "-"))
+	resourceName := "lxd_profile.profile1"
+
+	resource.Test(t, resource.TestCase{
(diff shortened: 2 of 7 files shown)


mjrider:
not sure why the storagepool tests fail, i did not change anything in the storagepools

jtopjian:
@mjrider It's nothing you did. It's either a Travis issue or upstream LXD issue.

mjrider:
seems to be an lxd daemon issue........ see lxc/lxd#3877
strange thing, i cannot reproduce it locally, only on travis. see https://github.com/mjrider/terraform-provider-lxd/tree/test-build for my test

jtopjian:
Thanks for looking into this. I'm not entirely convinced it's an lxd daemon issue yet. I have an open PR in which I've been trying various things.

One thing to note is that Travis does not support Ubuntu 16.04. All tests are being run in 14.04, regardless of specifying 16.04. I have been able to get some tests to fail in 14.04 outside of Travis. So it could be a kernel issue or something like that.

It might be worthwhile to postpone opening a bug report with lxd until the same failures can be reproduced outside of both Terraform and Travis.

Another option is to try older lxd releases to see if the problem goes away. If it does, you could look at commits and see what might have triggered this issue and then provide further information to the lxd team.

mjrider:
@jtopjian something like this ? https://github.com/mjrider/lxd-setup-test

jtopjian:
Yep, exactly.

jtopjian:
I think I've narrowed it down to this commit: https://github.com/lxc/lxd/commit/ac851c1f37263ea4497edf028a23747b7879485e

This is only happening if the storage pool dir is created outside of `/var/lib/lxd/storage-pools`.

If it's created outside, then the `source` dir is being removed, but the bind-mounted dir at `/var/lib/lxd/storage-pools/foo` is triggering the `device or resource busy` error.

jtopjian:
Getting closer :)

Manually running `umount /tmp/foo` allows the `storage pool delete` command to run successfully.

mjrider:
maybe usefull to tell you findings in the lxd issue aswell

jtopjian:
Yep - I did shortly after posting the above comment :)

mjrider:
sorry, totaly missed that part, my brain needs more coffee

jtopjian:
No problem at all :)

mjrider:
finally we have a successful build again, so this is ready for merge imho

jtopjian:
@mjrider Thanks for submitting this. Is there a requirement to have all of the ubuntu images updated to `ubuntu:xenial`? Most of the images used for testing are now using `alpine` to reduce resources used in the test suite and I've been looking to get rid of the remaining "ubuntu" ones at some point.

If `ubuntu:xenial` helps trigger the scenario for this test, then it's completely fine to keep it.

mjrider:
@jtopjian nope, that is leftover from debugging, will fix that later today

jtopjian:
> brainwave few minutes later, this makes the test work without copying a remote image to locale with the alias ubuntu, i can change them to alpine if you would like that

It's really six-one-way-half-a-dozen-the-other :)

wrt the local image - correct, we are downloading the ubuntu image in Travis. Ideally that should be changed to alpine, too. It's probably worthwhile to just bundle all of that in a separate, dedicated PR.

I'll go ahead and merge this as-is.

jtopjian:
Oh - nevermind - just saw you changed all the images to alpine. Perfect. I'll do another PR soon to clean the rest up.

PR #98, opened 2017-09-29
wip

jtopjian:


jtopjian:
I'm starting to think this is an Ubuntu 14.04 issue. I have not been able to reproduce this under 16.04, but I created a 14.04 test environment and the suite of tests failed to finish.

Unfortunately it looks like Travis does not support Ubuntu 16.04. Even though I specified "xenial" in this PR, 14.04 is still being used. There are open issues at the Travis github org that mention 16.04 is not yet available.

PR #99, opened 2017-10-01
add import for storage pools wth basic import test

mjrider:
add import support for storage pools

add the following to the resource config
```hcl
resource "lxd_storage_pool" "default" {
  name = "default"
  config {
    source = "/var/lib/lxd/storage-pools/default"
  }
  driver = "dir"

}
```

and run the command
`terraform import lxd_storage_pool.default default`

PS: i created for the storage pool test issues the issue lxc/lxd#3877, because i can reproduce them with lxc cli command

### DIFF
--- a/lxd/resource_lxd_storage_pool.go
+++ b/lxd/resource_lxd_storage_pool.go
@@ -3,6 +3,7 @@ package lxd
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/lxc/lxd/shared/api"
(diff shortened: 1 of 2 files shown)


lxd/resource_lxd_storage_pool.go
```diff
@@ -161,3 +174,32 @@ func resourceLxdStoragePoolExists(d *schema.ResourceData, meta interface{}) (exi
 
 	return
 }
+
+func resourceLxdStoragePoolImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	p := meta.(*LxdProvider)
+	remote, name, err := p.Config.ParseRemote(d.Id())
+	log.Printf("[DEBUG] Import storage pool from remote: %s name: %s", remote, name)
```
jtopjian: This should be moved to after the `if err != nil` check for safety.

lxd/resource_lxd_storage_pool.go
```diff
@@ -161,3 +174,32 @@ func resourceLxdStoragePoolExists(d *schema.ResourceData, meta interface{}) (exi
...
+		d.Set("remote", remote)
+	}
+
+	server, err := p.GetContainerServer(p.selectRemote(d))
+	if err != nil {
+		return nil, err
+	}
+
+	pool, _, err := server.GetStoragePool(name)
+	log.Printf("[DEBUG] Import Retrieved storage pool %s: %#v", name, pool)
```
jtopjian: Same here.

jtopjian:
@mjrider Looks good - thanks!

PR #100, opened 2017-10-01
switch to snap for installing lxd

mjrider:
Switched the ppa out for using snap to install lxd. 
This seems to work, while the ppa doesn't

not sure if you want this change, but i was looking for a way to test my PR's

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,12 @@ env:
    LXD_PASSWORD=the-password \
 
 before_install:
-  - sudo add-apt-repository -y ppa:ubuntu-lxc/lxd-stable
-  - sudo apt-get -qq update
-  - sudo apt-get install -y lxd
+  - sudo apt-get install -y snapd
+  - sudo snap install lxd
+  - export PATH=/snap/bin/:${PATH}
+  - sleep 10 # give lxd time to start
   - sudo lxd init --auto --trust-password="${LXD_PASSWORD}" --network-port=8443 --network-address='[::]'
-  - sudo chmod 777 /var/lib/lxd/unix.socket
+  - sudo chmod 777 /var/snap/lxd/common/lxd/unix.socket
   - 'lxc image copy ubuntu:t local: --alias=ubuntu'
   - lxc image list
 


jtopjian:
@mjrider Nice find :) 

It's certainly something to fall back on. I'm going to dig into this a little more and see if I can pinpoint the problem. I'm open to using the snap method if I can't find anything, though.

jtopjian:
@sl1pm4t I'm thinking we should move forward with this. Given Travis's lack of 16.04 support, using the snap will ensure we're testing LXD with the latest intended supporting libraries. Thoughts?

sl1pm4t:
I have no problem with the change, but I'm also not familiar with snap beyond the basics, so I'll leave  up to you @jtopjian.

jtopjian:
I'm in the same boat. I have a cursory amount of knowledge with snap. _But_, I think this is relatively safe to do. If we start seeing complicated issues, we can just jump back to the ppa. :)

At a minimum, it'll fix Travis and let us start merging again.

jtopjian:
@mjrider Thank you for your investigation into this 😄

jtopjian:
Well, this is now failing in Travis and I have no idea why. I'm able to reproduce the error locally, too:

```
ubuntu@lxd-acc-tests:~$ sudo lxc ls                                                      
+------+-------+------+------+------+-----------+                                        
| NAME | STATE | IPV4 | IPV6 | TYPE | SNAPSHOTS |                                        
+------+-------+------+------+------+-----------+                                        
ubuntu@lxd-acc-tests:~$ lxc ls                                                           
chroot: cannot change root directory to '/var/lib/snapd/hostfs': Operation not permitted 
```

In the Travis log from the successful build, the only difference I can spot is this:

```
$ sudo snap install lxd
2017-10-01T19:13:57Z INFO Waiting for restart...
lxd 2.18 from 'canonical' installed
```

And the failing build has

```
$ sudo snap install lxd
2017-10-01T19:13:57Z INFO Waiting for restart...
lxd 2.18 from 'canonical' installed
```

Which might mean that the worker in which the Travis job ran recently had some updates that would not take effect until a reboot? 

This doesn't bode well for snapd being a viable alternative :/

We might have to switch back to the PPA and wait until LXD 2.19 is released.

mjrider:
looks we found a ubuntu/snap bug this time ......
https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1721234

jtopjian:
@mjrider Thanks for submitting the bug :)

> looking at the snap binary, i expected a suid root binary

ahh - that makes sense. I tried setting suid on the path `/var/lib/snapd/hostfs` but that didn't help. setuid on the binary makes sense.

mjrider:
except that making the binary suid, didn't help so i'm confused

jtopjian:
As you pointed out in the bug report, looks like this was another LXD bug, which was resolved :)

PR #103, opened 2017-10-05
Add initial import for containers

mjrider:
TODO:
- [x] figure out how to import the mandatory image variable, ideas are welcome
- [x] fix Read function to read all parameters
- [x] add documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,13 @@ resource "lxd_container" "test1" {
   }
 }
 ```
+Importing existing containers is possible with the following ID syntax
+terraform import lxd_container.container [remote:]name[/images:alpine/3.5].
+where:
+
+remote, optional, is the name of the remote in the provider config
+name, required, is the container name
+/images:alpine/3.5, optional, translates to image = images:alpine/3.5 in the resource configuration
 
 Note, the `config` attributes cannot be changed without destroying and re-creating the container, however values in `limits` can be changed on the fly.
 
--- a/README.md
+++ b/README.md
@@ -143,13 +143,6 @@ resource "lxd_container" "test1" {
   }
 }
 ```
-Importing existing containers is possible with the following ID syntax
-terraform import lxd_container.container [remote:]name[/images:alpine/3.5].
-where:
-
-remote, optional, is the name of the remote in the provider config
-name, required, is the container name
-/images:alpine/3.5, optional, translates to image = images:alpine/3.5 in the resource configuration
 
 Note, the `config` attributes cannot be changed without destroying and re-creating the container, however values in `limits` can be changed on the fly.
 
@@ -467,6 +460,14 @@ The following resources are currently available:
   * `file`      - *Optional* - File to upload to the container. See reference below.
   * `remote`    - *Optional* - The remote in which the resource will be created. If it
                                is not provided, the default provider remote is used.
+
+##### Importing
+Importing existing containers is possible with the following ID syntax  
+`terraform import lxd_container.container [remote:]name[/images:alpine/3.5]`.  
+
+ * remote             - *Optional* - is the name of the remote in the provider config
+ * name               - *Required* - is the container name
+ * /images:alpine/3.5 - *Optional* - translates to image = images:alpine/3.5 in the resource configuration
 
 ##### Container Network Access
 
--- a/lxd/import_resource_lxd_container_test.go
+++ b/lxd/import_resource_lxd_container_test.go
@@ -22,7 +22,6 @@ func TestAccContainer_importBasic(t *testing.T) {
 
 			resource.TestStep{
 				ResourceName: resourceName,
-				//				Config:            testAccContainer_basic(containerName),
 				ImportState:       true,
 				ImportStateVerify: true,
 				ImportStateId:     containerName + "/images:alpine/3.5/amd64",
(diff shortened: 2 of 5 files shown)


jtopjian:
This is a tough one.

The best case scenario is if we can find _something_ via the API that references the image. I did a cursory look (I can look more indepth if needed) and found a few references, but oddly nothing that references what we normally use in the `image` parameter.

The alternative is to pass in metadata via the command-line and parse it in the import function:

```shell
$ terraform import lxd_container.foo foo/ubuntu
```

Then do a split on `/` and use `ubuntu` as the image info.  This is a fairly common thing to do for importing... it's just kind of ugly.

mjrider:
@jtopjian so that would make the 'ID" syntax for example
remote:container/images:alpine/3.5/amd
i think that would work just fine
the options are then
- container
- remote:container
- container/image:description
- remote:container/image:description

and the parse rules:
- split in max 2 parts on /
- part1:
  - parse with lxd parse function like now
- part2:
  - set as image if is set


i already added the feature that in the special case that 'image' is set to "" (which happens now ) the diff functions ignores it, and terraform doesn't plan to destroy the container because of changes

jtopjian:
All sounds good to me.

Also, I just realized, we should update the README to document that importing exists for certain resources.

mjrider:
@jtopjian what do you think of the following documentation snipped

Importing existing containers is possible with the following ID syntax
terraform import lxd_container.container [remote:]name[/images:alpine/3.5].
where:
- remote, optional,  is the name of the remote in the provider config
- name, required,  is the container name
- /images:alpine/3.5, optional,  translates to `image = images:alpine/3.5` in the resource configuration

jtopjian:
@mjrider That sounds great.

mjrider:
@jtopjian @sl1pm4t did you have time to review this PR ?

README.md
```diff
@@ -143,6 +143,13 @@ resource "lxd_container" "test1" {
   }
 }
 ```
+Importing existing containers is possible with the following ID syntax
+terraform import lxd_container.container [remote:]name[/images:alpine/3.5].
+where:
+
+remote, optional, is the name of the remote in the provider config
+name, required, is the container name
+/images:alpine/3.5, optional, translates to image = images:alpine/3.5 in the resource configuration
```
jtopjian: I think this should be moved to the "Reference" area of the documentation rather than in the "Example" area. Perhaps a new section called "Importing" that is similar to [this](https://github.com/sl1pm4t/terraform-provider-lxd/blob/master/README.md#container-network-access).

Also, if you look at the rendered README, the last three lines are all placed on a single line. I think they should be formatted as a list in Markdown.

lxd/import_resource_lxd_container_test.go
```diff
@@ -0,0 +1,56 @@
...
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccContainer_basic(containerName),
+			},
+
+			resource.TestStep{
+				ResourceName: resourceName,
+				//				Config:            testAccContainer_basic(containerName),
```
jtopjian: nit: this looks weird.

lxd/resource_lxd_container.go
```diff
@@ -344,12 +348,28 @@ func resourceLxdContainerRead(d *schema.ResourceData, meta interface{}) error {
...
+		} else if strings.HasPrefix(k, "boot.") {
+			config[k] = v
+		} else if strings.HasPrefix(k, "environment.") {
+			config[k] = v
+		} else if strings.HasPrefix(k, "raw.") {
+			config[k] = v
+		} else if strings.HasPrefix(k, "security.") {
+			config[k] = v
+		} else if strings.HasPrefix(k, "user.") {
+			config[k] = v
```
jtopjian: Is it possible to use a `switch` statement here? It's usually cleaner to use a switch instead of multiple `else if` statements. If switch isn't possible, no problem.

lxd/resource_lxd_container.go
```diff
@@ -581,6 +601,43 @@ func resourceLxdContainerExists(d *schema.ResourceData, meta interface{}) (exist
...
+
+	ct, _, err := server.GetContainerState(name)
+	log.Printf("[DEBUG] Import container state %#v", ct)
+	log.Printf("[DEBUG] Import and the error %#v", err)
+	if err == nil && ct != nil {
+		d.SetId(name)
+		d.Set("name", name)
+	} else {
+		return nil, err
+	}
```
jtopjian: Could the above be rewritten as:

```go
ct, _, err := server.GetContainerState(name)
if err != nil {
    return nil, err
}

if ct == nil {
    return nil, fmt.Errorf("Unable to get container state")
}

log.Printf("[DEBUG] Import container state %#v", ct)
d.SetId(name)
d.Set("name", name)
```

lxd/resource_lxd_container.go
mjrider: afaik i can't do a switch with the hasPrefix call

jtopjian: Yes, you are absolutely correct.

jtopjian:
@mjrider Sorry for the delay.

Thank you for your work and patience with this one. It probably wasn't the easiest to implement, so your work is appreciated :)

PR #105, opened 2017-10-12
Provider Cleanup

jtopjian:
This commit performs some cleanup on the main provider code:

1. Removes depencency of lxdhelpers since the important functionality
is now in the upstream LXD/LXC API.

2. Adds some narrative comments to help explain all actions being
performed.

3. Rearranges some of the convenience methods and functions to keep
all methods together.

### DIFF
--- a/lxd/provider.go
+++ b/lxd/provider.go
@@ -1,6 +1,7 @@
 package lxd
 
 import (
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


jtopjian:
@sl1pm4t Back in August, I asked the LXD team to better expose the functions reproduced in the `lxdhelpers` package. In their usual helpful, awesome manner, they did within a matter of days. I'm just now getting around to integrating that functionality here. 

There is some added code from `lxdhelpers` here, but it's borderline negligible: saving a certificate to a file and the authentication function -- definitely not worth a whole separate package now.

The rest of the changes were me just making notes and rearranging some things as I re-learned how provider.go worked :)

PR #107, opened 2017-10-17
Add description field to network resource

### LINKED ISSUE #106: Description support
Hello!
Can u please add support for describing? For example, I want to set description to network interfaces like this:
resource "lxd_network" "lxd-internal" {
  name = "lxd-internal"
  description = "Internal network for containers on eth1"
  config {
    ipv4.address = "10.10.100.1/24"
    ipv4.nat     = "false"
    ipv6.address = "none"
    ipv6.nat     = "false"
  }
}

sl1pm4t:
Also removed Provider AccTest preCheck. It's no longer requried to run the tests on a system that has the lxc provider installed.

Fixes #106

### DIFF
--- a/lxd/provider_test.go
+++ b/lxd/provider_test.go
@@ -3,7 +3,6 @@ package lxd
 import (
 	"log"
 	"os"
-	"os/exec"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


jtopjian:
@sl1pm4t Not sure if you want this merged yet, so holding off.

sl1pm4t:
Thanks @jtopjian

PR #108, opened 2017-11-05
Fixed golint errors

mjrider:
After a bit of annoyance with my code editor (vscode) and the amount of lint errors reported by it, i decided that i wanted golint to be silent.

i tried to keep the commits focused on 1 error/type at the time so it is easy te revert changes which are considered unwanted. Most of the changes are made with gorename

summary:
- renamed Id to ID
- unexported symbols 
- rewrote for loops
- removed underscores from function names


jtopjian:
@mjrider Thanks for cleaning all of this up.

It was definitely a smart move to break each one out into a separate commit as I think some of these might have to be discussed. As an example: https://github.com/sl1pm4t/terraform-provider-lxd/pull/108/commits/d4854764d3950a04d5cffcc42a1b529348a60b02

While I understand that using underscores in function names isn't correct Go, this is used in Terraform exclusively for acceptance test names. You can see it in the AWS provider (which I consider to be the "reference" for Terraform providers):

https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/data_source_aws_ebs_volume_test.go

On the topic of `id/Id/ID`, the various forms are always a point of inconsistency. Terraform has functions `Id()` and `SetId()` which is why I think it's common to see `Id`. Personally I prefer `ID`. I'll defer this decision to @sl1pm4t :)

With regard to the editor you're using, I wonder if there's a way to ignore certain lint violations. 

I don't want the above comments to distract from any praise for doing this. It's always great to step back and clean things up. 

Another thing we could look at is adding tools like `goimports`, `go vet`, and possibly linting to the Travis gate. We use the first two tools at the gate in another project I work on and it really helps make sure code is clean and follows correct style.

mjrider:
@jtopjian thx for the comments,

the underscores was one of the 'uncertain' parts, if i check https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_vpn_connection_route_test.go
the test functions do not have a _ in the name. so even the aws 'reference' doesn't use them al the time.

for the Id vs ID parts, i think going with the 'golang prefers X'  option is the best way to go, muting lint errors is imho only a last resort when code cannot be changed or when the linter has false positives

sl1pm4t:
Hi @mjrider + @jtopjian 

Thanks for the big tidy up @mjrider. I also use VS Code and my linter has been yelling at me too!
I agree with all changes, and don't feel strongly about the underscore issue, but I lean on the side of keeping the underscores as @jtopjian says, to remain consistent with [most] of the other Terraform resources.

mjrider:
so to recap, if i revert d485476 this would be mergeable ?
because then i'll revert that commit so we can merge it
next on the list: adding qa tools to the travis configuration

mjrider:
@sl1pm4t + @jtopjian i removed the commit which removed the underscores so we can move  the discussion about golang vs terraform linting choices outside this PR and to a later moment

jtopjian:
@mjrider That sounds good to me. I've been busy with some other things and haven't had a chance to do a full review of this. I don't anticipate any more issues, but because I haven't devoted time to look at it in detail, I'm hesitant to merge it. If @sl1pm4t has and is happy with the changes, then I am too. :)

sl1pm4t:
Thanks @mjrider

PR #110, opened 2017-11-22
Updating test-infra files

jtopjian:
The latest LXD deb packages present a warning that the deb packages will be phased out toward the end of the year, so this updates the `deploy.sh` script to use the LXD snap and accounts for changes with the snap (such as volume handling). 

The Terraform files were automatically formatted with `hclfmt`.


test-infra/files/deploy.sh
```diff
@@ -5,34 +5,39 @@ vol=$(curl http://169.254.169.254/latest/meta-data/block-device-mapping/ebs0)
...
+_lxd="/snap/bin/lxd"
+sudo $_lxd waitready --timeout 60
+sudo $_lxc config set core.https_address [::]
+sudo $_lxc config set core.trust_password the-password
+#sudo $_lxc storage create default dir source=/mnt
+sudo $_lxc storage create default btrfs source=$vol
+sudo $_lxc profile device add default root disk path=/ pool=default
+sudo $_lxc network create lxdbr0 ipv6.address=none ipv4.address=192.168.244.1/24 ipv4.nat=true
+sudo $_lxc network attach-profile lxdbr0 default eth0
+sudo $_lxc image copy images:ubuntu/xenial/amd64 local: --alias ubuntu
```
mjrider: iirc i removed all the ubuntu images from the testcases, so maybe we can drop the ubuntu download

jtopjian: I agree and I'm OK with dropping it if it's necessary.

The reason Ubuntu is here for semi-selfish reasons. The `test-infra` stuff is what I use when I'm developing with LXD, so it's a mirror of my own environment. If I'm troubleshooting something, I'm much more familiar with Ubuntu than Alpine.

So the `test-infra` stuff is _kind of_ different than the Travis testing configuration. For Travis, we've standardized on Alpine for efficiency.

mjrider: you are not required to drop it , but i assumed it was an unintended leftover from when tests where done with ubuntu.

jtopjian: Yep, I understand. It's a completely legitimate thing to point out in the review. I totally understand how it can be mistaken as a leftover.

jtopjian:
@mjrider Got pulled away at work and forgot about this. Thanks for the review :)

Issue #82, opened 2017-08-23
file writes the sha1sum of the 'content' instead of the actual contents into the file


shantanugadgil:
Versions:
Submit Host: CentOS 7.3
Terraform: 0.10.2
terraform-provider-lxd: latest master built from sources
LXD Host: Ubuntu 16.04 + LXD 2.16 from xenial backports

Recently I have restarted by Terraform+LXD experiments with the latest versions and hit the following issue:

Many a times the contents of the file are the `sha1sum` of the content string rather than the content itself.

For the following file block:

```hcl
  file {
    content            = " "
    target_file        = "/etc/mysoftware/optional.conf"
    uid                = 0
    gid                = 0
    mode               = "0644"
    create_directories = true
  }
```

... what I get when I cat the file inside the container is:

```
[root@mylxdcontainer]# cat /etc/mysoftware/optional.conf
b858cb282617fb0956d960215c8e84d1ccf909c6
```

This happens to be the same as the following:
```
$ echo -n " " | sha1sum
b858cb282617fb0956d960215c8e84d1ccf909c6 *-
```

This happens even for blocks where the `content` is an interpolated output of ${file()}
This doesn't occur all the time, nor does it happen for all containers/files.
I haven't been able to figure out a pattern as to why this is occurring.

What I did is, I added a dummy file block above this "optional.conf" file block. (to mess up the ordering).
The subsequent plan detected a change in the file block count and the apply wrote out the correct file contents this time.

This behavior has me extremely puzzled.
I create about 8-10 LXD containers in one go, but I forcibly use `parallelism=2`.
I will try to repeat the exercise by doing a single container at one time and see if I can get predictably correct behavior.

Regards,
Shantanu Gadgil

jtopjian:
Yeah, that's _really_ bizarre behavior! Let me know if this issue goes away if you set `parallelism=1`.

shantanugadgil:
Hi @jtopjian , I have conducted my experiments during the day about 6-7 times now using `parallelism=1`.
I haven't hit an issue so far.

At first, I too thought that there was a random corruption, but then when I noticed that the strings were always fixed length, it led me to the sha1sum.
And then _that_ made a little bit more sense than it being a random corruption.

Thanks and Regards,
Shantanu

shantanugadgil:
Hi @jtopjian any thoughts why this would be happening when parallelism is **not** set to 1?

shantanugadgil:
[...]

PR #112, opened 2017-11-25
Container fixes

### LINKED ISSUE #53: high CPU consumption when using file block

### LINKED ISSUE #102: setting privileged=true has no effect
Hi,

Having experienced getting non-privileged containers even though I have set `privileged=true`, I think this is worth an issue here. Privileged contgainers can currently _only_ be obtained when `security.privileged = true` is set in the container's resource config, i.e.:

    resource "lxd_container" "my-container" {
      name = "my-container"
      image = "some-image"
      privileged = true     # this has no effect!

      config {
        security.privileged = "true"    # this puts the container in privileged mode
      }
    }

I haven't preformed extensive experiments (other than changing the config and seeing the error I experienced going away), but looking at the code in `resource_lxd_container.go`, I haven't found any use of the value of `privileged` that might go into the lxd api requests to create and start the container.

Even though the support for `privileged` might be necessary for backwards compatibility, it may make sense to do the following in order to avoid false expectations (and possibly some headache):

- adjust the documentation
- issue a warning when `privileged` is set with the hint to set `security.privileged` instead

jtopjian:
This PR contains two fixes for the `lxd_container` resource:

1. Deprecates the `privileged` parameter (fixes #102)
2. Adds the ability to specify a `source` for the `file` argument (fixes #53 and #82)

The implementation of 2 (notably the use of `homedir.Expand`) is based on the AWS S3 object resource.

2 is a breaking change. It will no longer store the checksum of the file contents in the state.

Additionally, there is no easy way to check and see if the content of the source file has changed. This means if someone updates the contents of the source file and runs `terraform apply`, the changes will not be noticed.

In my opinion, this is an acceptable limitation. Full management of files within the `lxd_container` resource is difficult. It would be much better to have a dedicated `lxd_container_file` resource (which I plan on creating shortly). I thought about completely removing the `lxd_container.file` argument entirely, but being able to add files _before_ the container starts has benefits. 

So I think the best practice should be:

* Use `lxd_container.file` for bootstrapping / prepping containers
* Use `lxd_container_file` for management of files in a running container

Once the `lxd_container_file` resource is made, I will detail this in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -462,8 +462,8 @@ The following resources are currently available:
                                is not provided, the default provider remote is used.
 
 ##### Importing
-Importing existing containers is possible with the following ID syntax  
-`terraform import lxd_container.container [remote:]name[/images:alpine/3.5]`.  
+Importing existing containers is possible with the following ID syntax
+`terraform import lxd_container.container [remote:]name[/images:alpine/3.5]`.
 
  * remote             - *Optional* - is the name of the remote in the provider config
  * name               - *Required* - is the container name
@@ -495,7 +495,8 @@ resource "lxd_container" "container1" {
 
 ##### File Block
 
-  * `content` - *Required* - The _contents_ of the file. Use the `file()` function to read in the content of a file from disk.
+  * `content` - *Required unless source is used* - The _contents_ of the file. Use the `file()` function to read in the content of a file from disk.
+  * `source` - *Required unless content is used* - The path to the source file to copy to the container.
   * `target_file` - *Required* - The absolute path of the file on the container, including the filename.
   * `uid` - *Optional* - The UID of the file. Must be an unquoted integer.
   * `gid` - *Optional* - The GID of the file. Must be an unquoted integer.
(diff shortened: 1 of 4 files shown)


jtopjian:
@mjrider Thank you for reviewing :)

I'm going to go ahead and merge. If anyone finds any issues with the changes to `lxd_container.file`, do let me know and I'll get them fixed up quickly.

PR #115, opened 2017-11-26
[WIP] LXD Provisioner

jtopjian:
This is the result of an afternoon of experimenting.

It works, but in order for it to work, the binary `terraform-provisioner-lxd` must exist. It's just an exact copy of `terraform-provider-lxd`. I'm sure even a symlink will work. I haven't found a way around this yet. It's possible there isn't a way around this.

Example:

```hcl
resource "lxd_container" "container1" {
  name = "foo"                         
  image = "images:alpine/3.5/amd64"    
  profiles = ["default"]               
                                       
  file {                               
    content = "Goodbye, World!\n"      
    target_file = "/foo/bar.txt"       
    mode = "0644"                      
    create_directories = true          
  }                                    
                                       
  provisioner "lxd" {                  
    inline = [                         
      "pwd",                           
      "ls",                            
      "echo hi",                       
      "echo $SHELL",                   
    ]
  } 
} 
```

The result will be the inline set of commands being run via `lxc exec`.

I'm not sure if this is something we'd want to merge? The dual-binary seems a little clunky. But it works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,8 @@ build:
 	tar czvf terraform-provider-lxd_${TRAVIS_TAG}_linux_amd64.tar.gz terraform-provider-lxd
 
 dev:
-	go build -v
+	go build -v -o terraform-provider-lxd
+	go build -v -o terraform-provisioner-lxd
 
 vet:
 	@echo "go vet ."
--- a/lxd/provider.go
+++ b/lxd/provider.go
@@ -25,6 +25,7 @@ type lxdProvider struct {
 
 	acceptRemoteCertificate bool
 	clientMap               map[string]lxd.Server
+	remoteSchemas           map[string]interface{}
 }
 
 // Provider returns a terraform.ResourceProvider
@@ -179,6 +180,7 @@ func init() {
 
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	var config *lxd_config.Config
+	remoteSchemas := make(map[string]interface{})
 
 	// Load remotes from LXC config
 	//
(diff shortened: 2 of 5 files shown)


shantanugadgil:
Having a provisioner will be an extremely welcome addition.
I personally don't mind having two copies of the same binary.

For my use case, I have to run some post install commands and having them inline would be easier to maintain than from an external script.

jtopjian:
I agree that a provisioner would be welcomed. 

But along with the two binaries, another thing I'm not a fan of is what needs to be done to get it to work with a `null_resource`. In order for that to work with this provisioner, you'd have to do the following:

```hcl
resource "null_resource" "provisioner" {
  connection {
    name = x
    remote = x
    address = x
    scheme = x
    password = x
    etc
  }
  
  provisioner "lxd" {
    inline = [ ... ]
  }
}
```

This provisioner works great when it's embedded in the `lxd_container` resource because it has automatic access to the container's connection info. But beyond that, the connection info needs specified. The provisioner has no access to the provider's information at this time.

Maybe there's a way around this or maybe Terraform will be updated in the future to enable better information sharing between providers and provisioners.

jtopjian:
I haven't forgotten about this one. I've been meaning to ask the Terraform core team about this for advice on if there's a better approach.

segmond:
Just wanted to mention that it's possible to provision as is with a little bit of creativity.

\# setup ssh so we can ssh in
  provisioner "local-exec" {
    command = "lxc exec -- test${count.index} apt update -y"
    command = "lxc exec -- test${count.index} apt install openssh-server -y"
    command = "lxc file push terraform.pub test${count.index}/root/terraform.pub ; lxc exec --verbose -- test${count.index} sh -c \"sleep 5; sudo mv /root/terraform.pub /home/ubuntu/.ssh/authorized_keys\""
  }

\# push up a script if you have a large script to run
  provisioner "file" {
    connection {
      type = "ssh"
      user = "ubuntu"
      private_key = "${file("/home/segmond/Work/terraform/terraform.private")}"
    }
    source = "script/script1.sh"
    destination = "script1.sh"
  }

\# provision with remote-exec
  provisioner "remote-exec" {
    connection {
      type = "ssh"
      user = "ubuntu"
      private_key = "${file("/home/segmond/Work/terraform/terraform.private")}"
    }
    inline = [
      "touch xyz123",
      "pwd > /tmp/pwd.path",
      "env > /tmp/env.data",
      "chmod 700 /home/ubuntu/script1.sh",
      "sudo /home/ubuntu/script1.sh"
    ]
  }

jtopjian:
> Just wanted to mention that it's possible to provision as is with a little bit of creativity.

Right, this is completely possible. It's also possible to continue using a `local-exec` provisioner and shelling out to the `lxc` command.

But what this PR does is use the LXD/LXC's native API for command execution. There are two benefits to this:

1. You don't need to have the LXC command installed locally.
2. You don't need to have an SSH server running on all of your containers (meaning, if you had 50 containers running on your host, you wouldn't have 50 `sshd` daemons running.

jtopjian:
I had a really good chat about this last week.

The way this is implemented is the only good way at this time. This implementation is something that has been identified as useful for a few other providers (for example: Docker). 

There are future plans to more closely align Providers and Provisioners so that Provisioners can easily read data from Providers. There's no time line for this support, though.

I'm hesitant to merge this since this might require a large refactor in the future. We can still leave it open in case there are some users who want to do a custom compile.

inercia:
Hi guys, any plans on merging this in the near future? I think it would be really good to be able to run commands this way as it would allow us to use standard images...

jtopjian:
Four years goes by pretty quick...

I originally left this PR open for reference or if anyone wanted to use this locally since this was more of an experiment rather than an official feature. There are probably too many API fixes that need to go into making this work, so I am going to close this.

PR #114, opened 2017-11-25
Docs Refactor

jtopjian:
For #111 

You can view the rendered documentation here: https://github.com/jtopjian/terraform-provider-lxd/tree/docs

This follows conventions of the core Terraform providers documentation format.

PR #116, opened 2017-11-26
Container file resource

jtopjian:
This commit adds the lxd_container_file resource which can be used
to manage individual files in a container.

This depends on #112 

I can rebase afterwards or all 3 commits can be merged here.

Can also add docs once #114 is added.

### DIFF
--- a/lxd/shared.go
+++ b/lxd/shared.go
@@ -239,7 +239,7 @@ func containerUploadFile(server lxd.ContainerServer, container string, file File
 func containerDeleteFile(server lxd.ContainerServer, container string, targetFile string) error {
 	targetFile, err := filepath.Abs(targetFile)
 	if err != nil {
-		return fmt.Errorf("Could not santize destination target %s", targetFile)
+		return fmt.Errorf("Could not sanitize destination target %s", targetFile)
 	}
 
 	targetIsDir := strings.HasSuffix(targetFile, "/")
--- a/lxd/shared.go
+++ b/lxd/shared.go
@@ -3,7 +3,14 @@ package lxd
 import (
 	"fmt"
 	"log"
+	"os"
+	"path"
+	"path/filepath"
+	"strconv"
 	"strings"
+
+	lxd "github.com/lxc/lxd/client"
+	"github.com/mitchellh/go-homedir"
 )
 
 // Complex resource ID types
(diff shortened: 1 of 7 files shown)


mjrider:
I'm wondering why you want to move the file subresource to a seperate resource
from the following 2 examples, i think the second is the most readable, and has the least surprises, so which scenario are you trying to cover here?

```hcl
resource "lxd_container" "container1" {
   name = "foobar"
   image = "images:alpine/3.5"
   profiles = ["default"]
 }
 
 resource "lxd_container_file" "file1" {
   container_name = "${lxd_container.container1.name}"
   remote  = "${lxd_container.container1.remote}"
   target_file = "/foo/bar.txt"
   content = "Hello, World!\n"
   create_directories = true
 }
```

vs
```hcl
resource "lxd_container" "container1" {
   name = "foobar"
   image = "images:alpine/3.5/amd64"
   profiles = ["default"]
   file {
     target_file = "/foo/bar.txt"
     content = "Hello, World!\n"
     create_directories = true
   }
 }
```

jtopjian:
@mjrider See my comments in https://github.com/sl1pm4t/terraform-provider-lxd/pull/112

sl1pm4t:
What would be the problem you see with keeping the original embedded file sub resource along side the standalone resource?
 I know it would some caveats ( the two resources compete and TF will see a diff on each run  ) but it seems to be a commonly accepted pattern in other providers. e.g google_container_cluster that includes a node_pool sub resource vs the standalone google_container_node_pool resource.

jtopjian:
Rebased and docs added.

mjrider:
@jtopjian what do you think about  sl1pm4t suggestion for keeping the subresource version , so users can have a choice in what they need/want

jtopjian:
@mjrider That was my intention all along :smile:

shantanugadgil:
FWIW, I don't mind having _only_ a separate resource.

lxd/shared.go
```diff
@@ -127,3 +155,155 @@ func resourceLxdValidateDeviceType(v interface{}, k string) (ws []string, errors
...
+	log.Printf("[DEBUG] Successfully uploaded file %s", targetFile)
+
+	return nil
+}
+
+// containerDeleteFile will delete a file on a container.
+func containerDeleteFile(server lxd.ContainerServer, container string, targetFile string) error {
+	targetFile, err := filepath.Abs(targetFile)
+	if err != nil {
+		return fmt.Errorf("Could not santize destination target %s", targetFile)
```
sl1pm4t: typo - `santize` -> `sanitize`

sl1pm4t:
>  That was my intention all along 😄

Sorry for the confusion, when I read through the PR comments on my phone I somehow got the impression you were removing the embedded resource. Might be because there was another PR for deprecated a resource, and I confused the two.

Anyway it looks good to me.

jtopjian:
No problem at all!

I detailed my plans for this in https://github.com/sl1pm4t/terraform-provider-lxd/pull/112. I should have added those notes here as well, so it's my fault for the confusion.

Thank you for catching the typo.

PR #117, opened 2017-11-26
Volume Profile Attach Resource

jtopjian:
This commit adds the lxd_volume_profile_attach resource which allows
attaching a volume to a profile.

### DIFF
--- a/lxd/provider.go
+++ b/lxd/provider.go
@@ -155,6 +155,7 @@ func Provider() terraform.ResourceProvider {
 			"lxd_storage_pool":            resourceLxdStoragePool(),
 			"lxd_volume":                  resourceLxdVolume(),
 			"lxd_volume_container_attach": resourceLxdVolumeContainerAttach(),
+			"lxd_volume_profile_attach":   resourceLxdVolumeProfileAttach(),
 		},
 
 		ConfigureFunc: providerConfigure,
--- a/lxd/resource_lxd_profile.go
+++ b/lxd/resource_lxd_profile.go
@@ -1,6 +1,7 @@
 package lxd
 
 import (
+	"fmt"
 	"log"
 
 	"github.com/hashicorp/terraform/helper/schema"
@@ -34,6 +35,7 @@ func resourceLxdProfile() *schema.Resource {
 			"device": &schema.Schema{
 				Type:     schema.TypeList,
 				Optional: true,
+				Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"name": &schema.Schema{
@@ -125,7 +127,12 @@ func resourceLxdProfileRead(d *schema.ResourceData, meta interface{}) error {
 		device["properties"] = lxddevice
 		devices = append(devices, device)
 	}
-	d.Set("device", devices)
+
+	err = d.Set("device", devices)
+	if err != nil {
+		return fmt.Errorf("unable to set profile %s devices: %s", name, err)
+	}
+
 	return nil
 }
 
--- a/lxd/resource_lxd_volume_profile_attach.go
+++ b/lxd/resource_lxd_volume_profile_attach.go
@@ -1,0 +1,217 @@
+package lxd
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
+	lxd "github.com/lxc/lxd/client"
+)
+
+func resourceLxdVolumeProfileAttach() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceLxdVolumeProfileAttachCreate,
+		Delete: resourceLxdVolumeProfileAttachDelete,
+		Exists: resourceLxdVolumeProfileAttachExists,
+		Read:   resourceLxdVolumeProfileAttachRead,
+
+		Schema: map[string]*schema.Schema{
+			"pool": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+
+			"volume_name": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+
+			"profile_name": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+
+			"path": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+
+			"device_name": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Optional: true,
+				Computed: true,
+			},
+
+			"remote": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func resourceLxdVolumeProfileAttachCreate(d *schema.ResourceData, meta interface{}) error {
+	p := meta.(*lxdProvider)
+	server, err := p.GetContainerServer(p.selectRemote(d))
+	if err != nil {
+		return err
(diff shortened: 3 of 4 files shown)


jtopjian:
Looks like there's a test failure. I'll look into this.

mjrider:
i found an issue with this, not sure if it's related to the testfail

3 runs of terraform apply
```
muller@sleeptop ~/gocode/src/github.com/sl1pm4t/terraform-provider-lxd/test $ terraform apply 
lxd_storage_pool.pool1: Creating...
  config.%:      "" => "1"
  config.source: "" => "/tmp/foo"
  driver:        "" => "dir"
  name:          "" => "foo"
lxd_profile.profile1: Creating...
  device.#:                    "" => "1"
  device.0.name:               "" => "eth0"
  device.0.properties.%:       "" => "1"
  device.0.properties.nictype: "" => "p2p"
  device.0.type:               "" => "nic"
  name:                        "" => "foo"
lxd_storage_pool.pool1: Creation complete after 0s (ID: foo)
lxd_profile.profile1: Creation complete after 0s (ID: foo)
lxd_volume.volume1: Creating...
  expanded_config.%: "" => "<computed>"
  name:              "" => "foo"
  pool:              "" => "foo"
  type:              "" => "custom"
lxd_volume.volume1: Creation complete after 0s (ID: foo/foo/custom)
lxd_volume_profile_attach.attach1: Creating...
  device_name:  "" => "foo"
  path:         "" => "/tmp"
  pool:         "" => "foo"
  profile_name: "" => "foo"
  volume_name:  "" => "foo"
lxd_volume_profile_attach.attach1: Creation complete after 0s (ID: foo/foo/foo)

Apply complete! Resources: 4 added, 0 changed, 0 destroyed.
muller@sleeptop ~/gocode/src/github.com/sl1pm4t/terraform-provider-lxd/test $ terraform apply 
lxd_storage_pool.pool1: Refreshing state... (ID: foo)
lxd_profile.profile1: Refreshing state... (ID: foo)
lxd_volume.volume1: Refreshing state... (ID: foo/foo/custom)
lxd_volume_profile_attach.attach1: Refreshing state... (ID: foo/foo/foo)
lxd_profile.profile1: Modifying... (ID: foo)
  device.#:                   "2" => "1"
  device.1.name:              "foo" => ""
  device.1.properties.%:      "3" => "0"
  device.1.properties.path:   "/tmp" => ""
  device.1.properties.pool:   "foo" => ""
  device.1.properties.source: "foo" => ""
  device.1.type:              "disk" => ""
lxd_profile.profile1: Modifications complete after 0s (ID: foo)

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
muller@sleeptop ~/gocode/src/github.com/sl1pm4t/terraform-provider-lxd/test $ terraform apply 
lxd_storage_pool.pool1: Refreshing state... (ID: foo)
lxd_profile.profile1: Refreshing state... (ID: foo)
lxd_volume.volume1: Refreshing state... (ID: foo/foo/custom)
lxd_volume_profile_attach.attach1: Refreshing state... (ID: foo/foo/foo)
Error refreshing state: 1 error(s) occurred:

* lxd_volume_profile_attach.attach1: 1 error(s) occurred:

* lxd_volume_profile_attach.attach1: lxd_volume_profile_attach.attach1: Unable to determine device name for volume foo on profile foo
```

config:
```hcl
provider "lxd" {}

resource "lxd_storage_pool" "pool1" {
  name = "foo"
  driver = "dir"
  config {
    source = "/tmp/foo"
  }
}

resource "lxd_volume" "volume1" {
  name = "foo"
  pool = "${lxd_storage_pool.pool1.name}"
}

resource "lxd_profile" "profile1" {
  name = "foo"
  device {
    name = "eth0"
    type = "nic"
    properties {
     nictype ="p2p"
    }  
  }
}

resource "lxd_volume_profile_attach" "attach1" {
  pool = "${lxd_storage_pool.pool1.name}"
  volume_name = "${lxd_volume.volume1.name}"
  profile_name = "${lxd_profile.profile1.name}"
  path = "/tmp"
  device_name = "foo"
}
```


without the nic in the profile, it all works as expected

jtopjian:
ahh, good catch. I'll look into this, too.

mjrider:
dunno how to make a testcase to test the whole 'run this 3 times' scenario, otherwise you would have got a patch for that, but if we find out, the container_attach needs the same test imho

mjrider:
did some testing, adding
```hcl
  device {
    name = "eth0"
    type = "nic"
    properties {
     nictype ="p2p"
    }  
  }
```

to the _basic test is enough to trigger a test failure for this case

jtopjian:
OK, cool. So it looks like it's an issue with device detection when there is more than one device. Thank you for looking into this.

jtopjian:
I see what's going on here. It's Terraform being Terraform :)

To understand the problem, think about how Terraform is building the resources:

1. Create a Volume
2. Create a Profile with a NIC device
3. Attach the Volume to the Profile, which adds a new device to the profile.

In Step 2, a device is specified in the Terraform configuration. But in Step 3, a new device is added to the LXD profile. When Terraform goes to refresh the profile, it sees 1 device in the `.tf` config, but the LXD API is saying there are 2.

There's no easy way of remedying this. Terraform cannot merge what both the `.tf` config defines and what the API reports. This is fairly common problem with building Terraform resources.

The decision of how to handle this might have long-term implications, so let's take a step back and look at why I created the `lxd_volume_container_attach` and `lxd_volume_profile_attach` resources: simply to emulate what the command-line provides:

```shell
$ lxc storage | grep attach                                                                
lxc storage volume attach [<remote>:]<pool> <volume> <container> [device name] <path>      
lxc storage volume attach-profile [<remote:>]<pool> <volume> <profile> [device name] <path>
```

This might have been a mistake since emulating what is done on the command-line does not always translate well to Terraform.

So what are `lxc storage volume attach` and `attach-profile` doing behind the scenes? The same thing as `lxc profile device add`. Which means it's no different than specifying a `device` block in a container or profile. So the functionality to attach volumes already exists in Terraform.

If we choose to keep these attach resources, that means we choose to use the LXC CLI as the model for which resources are made. If we choose to remove them, we choose to use an understanding of the LXD API and how Terraform can best consume it. There are pros and cons to both. However, I am in favor of the API.

Another thing to consider is that the `device` block in both the container and profile resource pose the same long-term problems as the `lxd_container_file` resource: while we might not immediately see problems specifying a `file` or `device` inline of the resource, from experience I can say that someone eventually will. However, I don't think it's critical to solve this now. It's always possible to wait until someone comes up with a valid use-case. 

So we have a few options:

1. Keep the attach resources. Modify the container and profile resources to have a `devices` block and an `all_devices` block. What this does is allows users to define devices in `devices` and the API to set all known devices in `all_devices`. `devices` is write-only and `all_devices` is read-only. This is a common solution to this problem in Terraform, though it is acknowledged as a workaround.

2. Get rid of the attach resources. Since attaching volumes can be done through the `device` resource, supporting a dedicated resource provides two ways of accomplishing the same goal and that might lead to maintenance overhead. At the moment, I am in favor of this.

3. Get rid of the attach resources, make a dedicated `lxd_container_device` and `lxd_profile_device` resource, implement the `all_devices` pattern described above. This can be considered down the road.

Thoughts?

mjrider:
thought a while about it, i think i would go for option 2 for profiles and containers.1 sounds like a kludge, which results in not able to remove devices.
option 3 is maybe a way to go later, but  without the workaround mentioned in 1. So that would take some more thinking and designing

jtopjian:
Busy week at work and this got lost in my inbox.

I agree with this. Just a side note that the workaround described in Option 1 will allow devices to be removed, but only devices managed by Terraform.

I'll go ahead and close this and begin deprecating the `lxd_volume_container_attach_resource`.

PR #119, opened 2017-12-18
Build provider for linux, windows & mac

sl1pm4t:
The makefile has been updated to build and package the LXD provider for darwin, linux and windows.
The build version (git tag) is also included in the binary filename, to suit the new(ish) Terraform provider versioning scheme.


jtopjian:
:heart: 

Thanks, Matt!

PR #127, opened 2018-04-11
Lazy Load Remotes

jtopjian:
This commit enables lazy loading of LXD remotes.
This is useful for when the LXD remote might not
exist yet during an apply.

This Works For Me, but I'm marking this as WIP until acceptance testing clears -- there might be a few bugs it uncovers.

With this patch in place, I can deploy an OpenStack instance, deploy LXD on it, and then provision a container in the new LXD environment all in one manifest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ terraform-provider-lxd.iml
 **/*.swp
 settings.json
 **/terraform.tfstate*
+**/*.terraform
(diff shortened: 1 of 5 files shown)


jtopjian:
_almost_ got this :)

In general it's working but fortunately Travis flagged a concurrent map write issue: https://travis-ci.org/sl1pm4t/terraform-provider-lxd/jobs/372328940

jtopjian:
@sl1pm4t Finally got this working. Let me know if you have any questions or want to discuss this.

jtopjian:
@sl1pm4t Thank you!

PR #130, opened 2018-05-01
Container File: Recreate when not found

jtopjian:
This commit will cause an lxd_container_file resource to
be recreated when it's not found.


sl1pm4t:
Thanks @jtopjian

PR #131, opened 2018-05-01
Fix concurrent map write bug

jtopjian:


jtopjian:
I intentionally omitted this before since `Config` is part of the LXD packages and I figured it would be handled there. Lesson learned :)

jtopjian:
@mjrider Thanks!

PR #132, opened 2018-05-05
Container File: Recreate when container not found

jtopjian:
This commit modifies the behavior of the lxd_container_file
resource so that it is marked as not existing when the
container itself does not exist.

This is useful in situations when the container was deleted
out of band of Terraform.

Similar to #130 but now accounts for when the _container_ doesn't exist.


jtopjian:
Thanks, Matt!

PR #134, opened 2018-07-05
Bugfix: retry resourceLxdContainerRead if "inet" is not found in container status

renonat:
I have found that when lxd is under heavier than normal load (for example creating 4+ containers at the same time), it takes longer for the network information to propagate into the container state.
This has manifested into the issue where the container status pulled from lxd does not yet have a value for "inet", but the provider will continue to try connecting through ssh with a blank ip_address.

Previously, status was pulled once as soon as the container was "Running", with no "inet" ip_address. But, lxd will transition the container to "Running", and then afterwards add the "inet" ip_address. Thus, sshIP would remain blank, and the provider would repeatedly attempt to ssh in and fail until timeout.

This fix will simply re-pull the container status from lxc until an sshIP address has been found, or 30 retries (15s) have passed.
- In my case, where remote-exec tries to access the container, another few retries allow "inet" ip_address to be successful propagated.
- In another case, where the sshIP address is not needed, the same occurs. A few retries occur, and the "inet" is propogated.
- Finally, if the container never generates an "inet" ip_address, this simply delays the resourceLxdContainerRead function from returning as expected by ~15s.

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -331,6 +331,20 @@ func resourceLxdContainerCreate(d *schema.ResourceData, meta interface{}) error 
 		return fmt.Errorf("Error waiting for container (%s) to become active: %s", name, err)
 	}
 
+	// Lxd will return "Running" even if the "inet" has not yet been set.
+	// wait until we see an "inet" ip_address before reading the state
+	networkConf := &resource.StateChangeConf{
+		Target:     []string{"OK"},
+		Refresh:    resourceLxdContainerWaitForNetwork(server, name),
+		Timeout:    3 * time.Minute,
+		Delay:      refreshInterval,
+		MinTimeout: 3 * time.Second,
+	}
+
+	if _, err = networkConf.WaitForState(); err != nil {
+		return fmt.Errorf("Error waiting for container (%s) network information: %s", name, err)
+	}
+
 	return resourceLxdContainerRead(d, meta)
 }
 
@@ -349,6 +363,12 @@ func resourceLxdContainerRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 	log.Printf("[DEBUG] Retrieved container %s: %#v", name, container)
+
+	state, _, err := server.GetContainerState(name)
+	if err != nil {
+		return err
+	}
+	log.Printf("[DEBUG] Retrieved container state %s:\n%#v", name, state)
 
 	d.Set("ephemeral", container.Ephemeral)
 	d.Set("privileged", false) // Create has no handling for it yet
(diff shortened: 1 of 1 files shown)


sidxz:
I would also vote for this problem.
This happens often with me and not just when lxd is under heavier than normal load.
If the DHCP server is slightly slow in assigning an IP, this is the exact same behavior observed. Happens almost 3 ~ 4 times out of 10.

renonat:
@jtopjian That makes sense. I'll do some more testing and try polling in the Create function.

renonat:
@jtopjian I have only found the problem to occur when the container is initially loading. 
While polling in the read function would be ideal, it would require repeating the code that parses inet from state.
Creating a new WaitForState function seems excessive, as [terraform's own implementation](https://github.com/hashicorp/terraform/blob/e5cc8af7f3984d120ff76357cc60892f6a82fc7e/helper/resource/state.go) is over 250 lines of code
So I reverted to polling only the minimum amount required.

jtopjian:
>  I have only found the problem to occur when the container is initially loading.

To confirm: do you mean when it's initially being created? If not, then in what other situations would the container be loading? 

I want to make sure I understand the situation here because this would be the first time I've come across something where Terraform needed to poll in the `Read` function. I'm not implying that it's wrong, though.

> Creating a new WaitForState function seems excessive, as terraform's own implementation is over 250 lines of code

You don't need to _reimplement_ `WaitForState` - you can call the function from within the resource. It's already being done in the block of code I mentioned above :)

I would prefer leveraging `WaitForState` since, as you mentioned, it's 250 lines of code which can safely handle polling a resource in a flexible way :)

In a very rough way, it would look something like this:

```go
stateConf := &resource.StateChangeConf{
  Target:     []string{"OK"},
  Refresh:    resourceLxdContainerWaitForNetwork(server, name),
  Timeout:    3 * time.Minute,
  Delay:      refreshInterval,
  MinTimeout: 3 * time.Second,
}

if _, err = stateConf.WaitForState(); err != nil {
  return fmt.Errorf("Error waiting for container (%s) network information: %s", name, 
}

// snip

func resourceLxdContainerWaitForNetwork(server lxd.ContainerServer, name string) resource.StateRefreshFunc {
  return func() (interface{}, string, error) {
    st, _, err := server.GetContainerState(name)
      if err != nil {
        return st, "Error", err
      }

      for iface, net := range state.Network {
        if iface != "lo" {
          for _, ip := range net.Addresses {
            // check if an IP exists, any IP.
            // if an IP is found:
            return st, "OK", nil
          }
        }
      }
    }      
}
```

Or something like that. Note that is just off the top of my head, has syntax errors, etc :)

But let me know if I am totally misunderstanding the problem here.

renonat:
That makes perfect sense. Thanks for your patience, this is my first time using Go and I totally misread how StateChangeConf worked.

jtopjian:
No problem at all! Let me know if you need any help or have any questions. I'll be able to do a more in-depth review later on.

jtopjian:
@renonat To confirm: this resolves the issue for you? I haven't run into this issue myself so it's hard for me to verify. I apologize for the obvious question, but want to make sure before I merge.

renonat:
@jtopjian Any idea when a release will be cut?

sl1pm4t:
Hey @renonat / @jtopjian I cut a new release today.
https://github.com/sl1pm4t/terraform-provider-lxd/releases/tag/v1.1.1

jtopjian:
Thanks, Matt!

PR #136, opened 2018-07-25
Provide clarity what happens when no lxd_remote has default = true

rjpearce:
Hopefully this is self-explanatory. I have tried to stick to the existing conventions but let me know if you want it tweaked.

I found this really confusing when I took the example provided and started using different remotes. I had to look at the source code to work out why an lxd_profile was only being created locally. 

Interesting side node, there is no mention of the lxd_remote(s) being used when you run terraform [plan|refresh|apply]. I might raise this as a separate issue.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -44,6 +44,7 @@ provider "lxd" {
     scheme   = "https"
     address  = "10.1.1.8"
     password = "password"
+    default  = true
   }
 
   lxd_remote {
@@ -86,6 +87,10 @@ The `lxd_remote` block supports:
 * `default` - *Optional* - Whether this should be the default remote. This
 	remote will then be used when one is not specified in a resource. Valid
 	values are `true` and `false`.
+  If you choose to _not_ set default=true on any `lxd_remote` this provider
+  will attempt to connect to an LXD server running on the same host through
+  the UNIX socket. See `Undefined Remote` for more information
+  The default can also be set with the `LXD_REMOTE` Environment variable.
 
 * `name` - *Optional* - The name of the LXD remote.
 


jtopjian:
@rjpearce Thank you for submitting this. I'm definitely in favor of anything that helps clarify usage.

Maybe we should further clarify the patch to read:

> If you choose to _not_ set default=true on an `lxd_remote` and do not specify a `remote` in a resource, this provider ...

And a slight (non-blocking) nit: could you indent the code to be flush with the rest of the existing `default` doc block?

rjpearce:
@jtopjian sure no problem. I have added the additional text suggested.

This is interesting because it mismatched with my current understanding. 
I had assumed that any resource without a remote would apply to all remotes, not the default remote. 
I will see if I can add further clarity to the docs to explain this in a new pull request. 

Sorry about the indentation (fixed), my vim configuration was hiding this from me. 

I've also noticed that the guidance under 'Undefined Remote' could do with some more clarity.
You can define an lxd_remote but it won't be used unless you make it default or have resources that specify this remote. Again I will see what I can do to make this more clear in a new pull request.

docs/README.md
```diff
@@ -84,13 +84,14 @@ The `lxd_remote` block supports:
...
-	remote will then be used when one is not specified in a resource. Valid
-	values are `true` and `false`.
-  If you choose to _not_ set default=true on any `lxd_remote` this provider
-  will attempt to connect to an LXD server running on the same host through
-  the UNIX socket. See `Undefined Remote` for more information
-  The default can also be set with the `LXD_REMOTE` Environment variable.
+* `default` - *Optional* - Whether this should be the default remote. 
+	This remote will then be used when one is not specified in a resource.
+	Valid values are `true` and `false`.
+	If you choose to _not_ set default=true on an `lxd_remote and do not specify
```
jtopjian: Small syntax error. Missing the closing backtick on `lxd_remote` 😄

rjpearce: Oops, sorry should be fixed now. :)

PR #138, opened 2018-08-25
profile device order

sergeyignatov:
golang's map is not ordered structure and terraform gets devices from lxd profile in random order.

### DIFF
--- a/lxd/resource_lxd_profile.go
+++ b/lxd/resource_lxd_profile.go
@@ -6,6 +6,7 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 
 	"github.com/lxc/lxd/shared/api"
+	"sort"
 )
 
 func resourceLxdProfile() *schema.Resource {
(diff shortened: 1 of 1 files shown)


jtopjian:
@sergeyignatov Thank you for reporting this and for providing a fix.

Would it be possible to supply an acceptance test for this that reproduces the issue and validates the fix?

sergeyignatov:
Go’s map iteration order (using the range keyword) is random instead of in the order that the entries were added. - here's the proof applying to lxd profiles https://play.golang.org/p/JpqP3_m-RjY

jtopjian:
@sergeyignatov What I meant was:

Can you add an acceptance test here: https://github.com/sl1pm4t/terraform-provider-lxd/blob/master/lxd/resource_lxd_profile_test.go

jtopjian:
@sergeyignatov Thank you for reporting this issue. I took a closer look at it and resolved it in #152. Please let me know if you have any questions.

PR #140, opened 2018-10-01
Build agains more target os/arch combinations

mjrider:
This change makes use of gox to compile more build targets. With some fixes on the side of lxd this should be able to build against almost all build targets available in gox.

Gox is using the multiple cpu cores available to speedup the build time

- Not building arm due to issues with compiling this plugin on travis
- build failures due to unrelated problem with travis building this plugin, a seperate PR to work arond that has been opened

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 GOFMT_FILES?=$$(find . -name '*.go' |grep -v vendor)
-TARGETS=darwin linux windows
+TARGETS=netbsd/386 freebsd/amd64 netbsd/arm linux/amd64 netbsd/amd64 openbsd/amd64 freebsd/386 darwin/amd64 windows/amd64 darwin/386 freebsd/arm openbsd/386 windows/386
 
 default: build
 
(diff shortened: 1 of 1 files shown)


sl1pm4t:
I agree @jtopjian, not much point building for architectures that Terraform doesn't target.
As for `gox`, I have no problem with this, my only hesitation is that the Makefile has lost targets to individually build a specific architecture. E.g. it's useful to be able to run `make darwin` and build a single arch for local testing. But I'm happy to let it go, if it's not part of your dev workflow.

mjrider:
i removed the os/arch combinations which are not needed

mjrider:
@jtopjian isn't that the same as this?
```
make targets TARGETS=darwin/amd64
gox -osarch='darwin/amd64' -output="dist/{{.OS}}_{{.Arch}}/terraform-provider-lxd__x4"
Number of parallel builds: 7

-->    darwin/amd64: github.com/sl1pm4t/terraform-provider-lxd
find dist -maxdepth 1 -mindepth 1 -type d -print0 | \
sed -z -e 's,^dist/,,' | \
xargs -0 --verbose --replace={} zip -r -j "dist/terraform-provider-lxd__{}.zip" "dist/{}"
zip -r -j dist/terraform-provider-lxd__darwin_amd64.zip dist/darwin_amd64 
  adding: terraform-provider-lxd__x4 (deflated 71%)
```

jtopjian:
@mjrider Yes, you're right :)

mjrider:
so it is ready for merge, or does it still need changes ?

jtopjian:
It looks good to me, but I'll let @sl1pm4t do the final approval since it alters releases and such.

jtopjian:
Regarding the Travis error that happened upon merge of this: I have a PR _almost_ ready to go. In addition to fixing some internal race conditions, I think I've discovered an upstream one, too. I'm going to open an Issue upstream to see if it's valid.

Issue #137, opened 2018-07-28
Possible Concurrent Map Write Bug


jtopjian:
Travis seems to be randomly flagging a concurrent map bug during the multi snapshot test.

This issue is to note the bug's existence. I should be able to look at it in the next few days.

shantanugadgil:
In the past when I have tried to create too many remote lxd containers with files (now lxd_file) within them, i used to hit many issues. When I made parallelism==1, it sort of mitigated the problem.

Weird, but deterministic things like the file's SHA1 being set as the file content.

This was quite a few months ago, so many things would have changed by now.

The word 'concurrent' for the lxd provider always makes me think of the issue I used to hit.

Just putting it out there, for a test case possibility.
I would recommend creating about 13-15 lxd containers in parallel (that was my use case then)

Cheers, Regards and Keep up the good work,
Shantanu
Hope this becomes a top-level provider!! 👍👍

PR #142, opened 2018-10-09
Cleaning up Locking

jtopjian:
This commit cleans up the locking done in the provider. It also
does another round of refactoring on the provider code by removing
unused code and renaming variables to be more clear.

For #137

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,10 @@ default: build
 
 test:
 	go get -d -t ./...
-	go test -timeout 20m -v ./lxd
+	go test -race -timeout 60m -v ./lxd
 
 testacc:
-	TF_LOG=debug TF_ACC=1 go test -v $(TESTARGS) -timeout 60m ./lxd
+	TF_LOG=debug TF_ACC=1 go test -v -race $(TESTARGS) -timeout 60m ./lxd
 
 build:
 	go build -v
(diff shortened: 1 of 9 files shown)


jtopjian:
There's a race error during the `TestAccContainer_importBasic` test which should be resolved with https://github.com/lxc/lxd/pull/5118. I'm marking this as `WIP` until that's resolved.

My goal is to get Travis to completely pass with `-race` enabled in the tests.

jtopjian:
I just can't win with this PR :smiley: 

Since Terraform has now merged the 0.12 breaking changes into master, and since 0.12 is still in heavy development, and since we don't version any libraries, Travis is going to fail for everything for the time being.

sl1pm4t:
Oh noes!
Maybe it's time to vendor TF 0.11 until 0.12 is stable?!

jtopjian:
Maybe. I'm going to review the errors in more detail when I have a chance. It's possible the changes required will be valid for 0.11 and 0.12. If that's the case, we'll be in good shape for when 0.12 is released.

If we have to vendor, maybe we can do something like a quasi-vendor where we have Travis check out a 0.11 branch before running the tests.

jtopjian:
hooray! 

Travis has run 9 times and no potential race conditions were detected. That's pretty good confidence that these changes are solid.

One note about the implementation: The `lxdProvider` contains an internal mutex while there's also a global mutex defined. This global mutex is for other areas of code which might require locking. An example can be seen with the network creation: https://github.com/sl1pm4t/terraform-provider-lxd/pull/142/files#diff-34facea7fe84ad143c8988f00c644bda

Issue #135, opened 2018-07-24
Issue with resourceLxdContainerWaitForNetwork


jtopjian:
The new `resourceLxdContainerWaitForNetwork` polling function works well. A little too well 😄 

There are some edge cases where the container won't have an IP address (whether no network, or network info is being supplied _after_ the container was created). In these cases, the apply will always fail.

I propose adding a new argument to `lxd_container` called `wait_for_network` with a default value of false. Users can opt-in to waiting for the network.

PR #143, opened 2018-10-09
#135: User opt-in to wait for network

renonat:
Feature request #135
Add `wait_for_network` argument to lxd_container resource. This argument allows `resourceLxdContainerWaitForNetwork` to be opt-in instead of always-on.

### DIFF
--- a/docs/resources/lxd_container.md
+++ b/docs/resources/lxd_container.md
@@ -80,6 +80,9 @@ resource "lxd_container" "container1" {
 
 * `file` - *Optional* - File to upload to the container. See reference below.
 
+* `wait_for_network` - *Optional* - Boolean indicating if the provider should wait for the container's network address to become available during creation.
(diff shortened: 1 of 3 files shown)


sl1pm4t:
My apologies @renonat  - I just saw @jtopjian's comment in #135 where he suggests the default value be `false`, so you had it correct first time around.

jtopjian:
I'm OK either way. 

After doing further testing, the chances of running a container with no network connectivity is less likely to happen than the latter. Coupled with retaining backwards compatibility, a default value of `true` makes sense.

jtopjian:
@renonat btw: thank you for implementing this. I've been meaning to get to this, but my cat-and-mouse game of hunting races/locks has kept me from it. :)

renonat:
@jtopjian Not a problem, after all I felt some responsibility for putting the original change in a few months ago

PR #144, opened 2018-10-15
update the terraform pluging path

pmalhaire:
It seems that the path for terraform plugin path changed since the README was writen those values worked for me.


jtopjian:
Confirmed: https://www.terraform.io/docs/configuration/providers.html#third-party-plugins

Thank you :)

PR #146, opened 2018-11-05
Testing: Pinning Terraform to v0.11.10 for now

jtopjian:
There's no way to get around the errors being reported when compiling/testing against Terraform v0.12. Some of the errors are reporting problems with HCL fixtures when there's no problem.

For now, I propose doing all tests against v0.11.10.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,10 @@ go:
   - master
 
 before_script:
-  -  go get -v -d -t ./...
+  - go get -v -d -t ./...
+  - cd $GOPATH/src/github.com/hashicorp/terraform
+  - git checkout v0.11.10
+  - cd -
 
 script:
   - make fmtcheck


shantanugadgil:
I will be sticking with 0.11.10 until 0.12.x launches with separate downloadable plugins. (Probably will wait till the .1 ;) )
Until then it is just a "test" version. :)

jtopjian:
@sl1pm4t Thanks!

@shantanugadgil From an end-user perspective, that's totally valid and probably safe. I know that the core team is putting _a lot_ of effort into making sure 0.12 is backwards compatible while still allowing all of the new HCL features. However, it's still quite a big change.

For developers, compiling against 0.12 is going to be mandatory for the plugin to work with 0.12 and prior. I was a little hesitant to pin testing to 0.11 because of this, but it looks as though the Terraform master branch isn't in a state to try yet.

Issue #133, opened 2018-06-20
Is this plugin tested with LXD 3.1?


HSBawa:
I just migrated to LXD 3.1. When I run my terraform script, I get following error on both Artful and Bionic.
```
* lxd_profile.icp_ce: Get http://unix.socket/1.0: dial unix /var/lib/lxd/unix.socket: connect: no such file or directory
* lxd_network.icp_ce_network: Get http://unix.socket/1.0: dial unix /var/lib/lxd/unix.socket: connect: no such file or directory
```
Not sure if this is a plugin or LXD issue. If I create following profile and network manually, with same parameters, they work fine.  Also, same code runs fine with LXD 2.21.

Thoughts?

jtopjian:
@HSBawa You might need to specify a different path to the `unix.socket`. The LXD snap is not located at `/var/lib/lxd/unix.socket`. Alternatively, you could symlink the snap path to `/var/lib/lxd/unix.socket` or communicate via https.

HSBawa:
@jtopjian  Thanks for quick response.You are correct. New location is: /var/snap/lxd/common/lxd/unix.socket. I will try making some changes and see if I can get it working using new path or else switch to https.

jtopjian:
@HSBawa No problem. I'm going to leave this issue open for now. Given all new versions of LXD will use `/var/snap/lxd/common/lxd/unix.socket` we should consider supporting both paths somehow.

HSBawa:
Using https, I was able to get past the Profile and Network. But for some reason container creation is still giving same issue.
```
provider "lxd"{
    generate_client_certificates = true
    accept_remote_certificate    = true
    ## format ex: 10s, 1m, 5h
    refresh_interval = "10s"

    #### Local access via "https"
    #### Workaround for snap lxd 
    lxd_remote {
        default  = true
        name     = "local-snap-lxd-1"
        scheme   = "https"
        address  = "127.0.0.1"
        port     = "9443"
        password = "password"
    }
}

```

```
* lxd_container.icp_ce_worker_2: Get http://unix.socket/1.0: dial unix /var/lib/lxd/unix.socket: connect: no such file or directory
* module.container.lxd_container.icp_ce_proxy: 1 error(s) occurred:

```

HSBawa:
Using https, I was able to get past the Profile and Network. But for some reason container creation was still giving same issue. 
```
[...]

PR #147, opened 2018-11-13
Support the new snap-based unix socket

jtopjian:
For #133 

This is a first attempt at transparently supporting the new snap-based socket path. It's a total hack which relies on the LXD client looking for the [`LXD_SOCKET`](https://github.com/lxc/lxd/blob/a8b6163f08ba18cc99ede3bc8ed4a7d141d2669b/client/connection.go#L88) environment variable. But the benefit should be that that this seamlessly works for the user.

If this is too much of a hack, totally understandable, and I'll try to come up with another method.


sl1pm4t:
@jtopjian let me know if you want me to cut a release after this is merged?

PR #150, opened 2018-12-12
Adding New Devices

jtopjian:
This commit adds two new device types as a valid device:
infiniband and proxy.


jtopjian:
Thanks, Matt!

PR #152, opened 2018-12-13
Change Devices from TypeList to TypeSet

jtopjian:
I was reviewing #138 and discovered that LXD does not report devices in any order. Sorting the results will only help if the user has defined their devices already sorted.

Unfortunately we will need to change devices from `TypeList` to `TypeSet`. If users are referencing devices using an index-based notation, this is going to break those references.

Commit message follows:

LXD does not return a container or profile's devices in a
set order. This causes random, sometimes perpetual, state
changes to occur.

The only way to resolve this issue is by changing the schema
type of devices from TypeList to TypeSet.

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -56,7 +56,7 @@ func resourceLxdContainer() *schema.Resource {
 			},
 
 			"device": &schema.Schema{
-				Type:     schema.TypeList,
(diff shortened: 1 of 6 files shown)


jtopjian:
Thanks, Matt!

jtopjian:
I don't foresee any other work at the moment - should we cut a new release?

PR #155, opened 2019-02-25
Fix creation_date / created_at

jtopjian:
The LXD snapshot field CreationDate has been renamed to CreatedAt.
This commit changes the value of the creation_date attribute to
use the new CreatedAt field. It also adds a new created_at attribute
and deprecates the older creation_date attribute.


jtopjian:
@sl1pm4t Thanks, Matt!

I'm hoping to have some time to get this ready for v0.12 shortly.

PR #157, opened 2019-03-20
Cleaning function names to avoid linter warnings

puppetmaster:
Not that important ... but it's a start

### DIFF
--- a/lxd/import_resource_lxd_container_test.go
+++ b/lxd/import_resource_lxd_container_test.go
@@ -17,7 +17,7 @@ func TestAccContainer_importBasic(t *testing.T) {
 		Providers: testAccProviders,
 		Steps: []resource.TestStep{
 			resource.TestStep{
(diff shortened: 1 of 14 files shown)


sl1pm4t:
Hi @puppetmaster, while I totally agree with this change, the unit test function names are currently following the naming convention used across all Terraform providers.
e.g. 
https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_instance_test.go

@jtopjian - do you have a preference?

jtopjian:
> while I totally agree with this change, the unit test function names are currently following the naming convention used across all Terraform providers.

This sums up my opinion as well. It's not specifically about the names, but making an effort to  follow the styles and methods of other Terraform providers.

sl1pm4t:
Closing as the current naming format following Terraform provider convention. Thanks anyway!

PR #159, opened 2019-04-03
vendor: switching to go modules

jtopjian:


jtopjian:
This should be good to go. 

I had to restrict the tests to only Go 1.12 because of checksum bugs in 1.11.

jtopjian:
Thanks, Matt!

PR #160, opened 2019-04-03
Terraform v0.12 Support

jtopjian:


jtopjian:
I'll rebase this at some point in the future.

We shouldn't merge this until v012 is ready. When that happens, we should probably do one final v011 release (if there are any changes that need to be published).

Or we could create a beta release based on this branch.

There are some syntax changes that are required for v012, but nothing major.

jtopjian:
This should be ready to go.

As an aside, it's no longer required to create binaries for v0.11 and v0.12. The latest v0.12 core code can support both.

PR #163, opened 2019-06-17
Acc Tests: Disable second network in access interface test

jtopjian:
Due to an unidentified timing issue, sometimes an LXD network
is not being created in time for the LXD container to launch
and is causing an out-of-ordering error to happen with the tests.

For now, we'll remove the second network from the test so there's
only one network to compare for test output.


sl1pm4t:
Thanks for looking at this @jtopjian

jtopjian:
No problem. I'm sorry I couldn't find the root cause. Perhaps someone will run into this in production and will be able to more easily pinpoint it.

PR #165, opened 2019-06-28
Fix windows build issues

jtopjian:


jtopjian:
The LXD dependency was also updated... I'll work on any bugs that come from that.

sl1pm4t:
Thanks for doing this!

PR #173, opened 2019-11-08
Expose containers IPv6 address

jgraichen:
This exposes two new container properties, `ipv4_address` and `ipv6_address`. The first one contains the same IPv4 address as in `ip_address`. The second property exports the containers IPv6 address.

The IPv6 address with the shortest representation is chosen, this should favor usual static addresses (e.g. `2001:db8::1234`) over dynamic addresses. I'm not sure if and how this can be integrated into `wait_for_network`, right now it does not wait explicitly for any IPv6 address. I had no problems yet.

Example:

```tf
resource "lxd_container" "mail" {
  name     = "mail"
  image    = "ubuntu"
}

resource "dns_aaaa_record_set" "public_mail_example_org" {
  zone      = "example.org."
  name      = "mail"
  addresses = [lxd_container.mail.ipv6_address]
}
```

### DIFF
--- a/docs/resources/lxd_container.md
+++ b/docs/resources/lxd_container.md
@@ -117,8 +117,14 @@ The `file` block supports:
 
 The following attributes are exported:
 
-* `ip_address` - The IP Address of the container. See Container Network Access
+* `ip_address` - The IPv4 Address of the container. See Container Network Access
   for more details.
+
+* `ipv4_address` - The IPv4 Address of the container. See Container Network
+  Access for more details.
+
+* `ipv6_address` - The IPv6 Address of the container. See Container Network
+  Access for more details.
 
 * `mac_address` - The MAC address of the detected NIC. See Container Network
   Access for more details.
@@ -128,8 +134,8 @@ The following attributes are exported:
 ## Container Network Access
 
 If your container has multiple network interfaces, you can specify which one
-Terraform should report the IP address of. If you do not specify an interface,
-Terraform will use the _last_ one detected.
+Terraform should report the IP addresses of. If you do not specify an interface,
+Terraform will use the _last_ IPv4 address detected and the IPv6 address with the shortest text representation.
 
 To specify an interface, do the following:
 
--- a/docs/resources/lxd_container.md
+++ b/docs/resources/lxd_container.md
@@ -135,7 +135,7 @@ The following attributes are exported:
 
 If your container has multiple network interfaces, you can specify which one
 Terraform should report the IP addresses of. If you do not specify an interface,
-Terraform will use the _last_ IPv4 address detected and the IPv6 address with the shortest text representation.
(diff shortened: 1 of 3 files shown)


jtopjian:
@jgraichen Thank you for submitting this. I'll investigate the test failures - it's probably due to upstream changes we need to account for in other areas of this provider.

With regard to the way the IPv6 address is detected, I'll think about this a little. I understand the reasoning, but there are cases where this won't work. For example, I run LXD containers in production and leverage the "longer" IPv6 addresses. 

Some initial thoughts are to strip out known local IPv6 prefixes and/or have `ipv6_address` be a list of multiple detected addresses.

jgraichen:
@jtopjian Thanks for you work!

I had thought about extending `wait_for_network` but did not implement anything yet. Maybe this could also solve the issue of deciding what "network" to wait for, and to select only addresses from specific networks:

```tf
  wait_for_network {
    ipv4 = [
      "10.0.0.0/24" # If addresses are from different networks this might work
      "10.0.0.0-99" # Prefer specific addresses? (e.g. if DHCP leases are 100-255 only)
    ] 
    ipv6 = [
      "2001:db8::/64{128}", # Only choose an address with a subnet mask of /128
                            # but within the /64 network
      "{scope:link}", # Use link local only (maybe there are no global address)
      "{shortest}", # some internal named filter?
      "{smallest}", # numerically smallest?
      "{first}"
    ]
    # if there is no match "the" network is not "up" yet...
  } 
```

These are just some wild ideas. I have not yet found any satisfying solution to choose an address. LXD does not expose if an IPv6 address is `temporary` too. This makes choosing from multiple addresses in the same network even harder.

Returning a list would be possible (and a nice addition, e.g. `ip_addresses`) but just shifts the problem to the user. I'm not sure how comfortable filtering with terraform functions or an external script would be...

Maybe I will try the above with an external script...

jtopjian:
I think we should use the first detected IPv6 address (possibly ignoring the fe80 address). If that's not suitable for the user, then they can configure the container so that the IPv6 address they want is the only/first address detected. All other solutions end up becoming quite complicated.

We already provide a way for the user to specify which NIC is the "access" NIC (`user.access_interface`), so in theory, all addresses on that NIC should be accessible. Therefore, any IPv6 address detected should be accessible and able to be used.

jgraichen:
I've change the IPv6 address selection to work exactly like the IPv4 address selection but to favor global addresses if present. Link local addresses will only be returned if there is no global address.

I'd like to add `ipv4_addresses` and `ipv6_addresses` in another MR if that is fine. Unfortunately terraform seems to have no functions for selecting specific addresses from a list ...

jtopjian:
@jgraichen I'm really sorry for the delay with this. This looks good in general - it probably just needs another rebase. I can take this over and work out the rebase if you'd like.

jgraichen:
@jtopjian No problem. I just did the rebase.

Issue #171, opened 2019-08-17
Profiles always cause changes


mikescott:
I have the following profile:

```terraform
resource "lxd_profile" "small" {
	name = "small"
	
    config = {
    	"limits.cpu" = 1
		"limits.memory" = "512MB"
		"boot.autostart" = true
    }
	
	device {
		name = "root"
		type = "disk"
		
		properties = {
			"pool" = "default"
			"path" = "/"
			"size" = "5GB"
			"limits.max" = "10MB"
		}
	}
	
	device {
		name = "eth0"
		type = "nic"
		
		properties = {
			"name" = "eth0"
			"nictype" = "bridged"
			"parent" = "lxdbr0"
			"limits.max" = "100Mbit"
		}
	}
}
```

Every time I run terraform plan or terraform apply it thinks there are changes, even though the profile hasn't been modified:

```
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

lxd_profile.small: Refreshing state... [id=small]
lxd_container.test1: Refreshing state... [id=test1]

------------------------------------------------------------------------

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # lxd_profile.small will be updated in-place
  ~ resource "lxd_profile" "small" {
        config = {
            "boot.autostart" = "true"
            "limits.cpu"     = "1"
            "limits.memory"  = "512MB"
        }
        id     = "small"
        name   = "small"

      + device {
          + name       = "eth0"
          + properties = {
              + "limits.max" = "100Mbit"
              + "name"       = "eth0"
              + "nictype"    = "bridged"
              + "parent"     = "lxdbr0"
            }
          + type       = "nic"
        }
      - device {
          - name       = "eth0" -> null
          - properties = {
              - "limits.max" = "100Mbit"
              - "nictype"    = "bridged"
              - "parent"     = "lxdbr0"
            } -> null
          - type       = "nic" -> null
        }
        device {
            name       = "root"
            properties = {
                "limits.max" = "10MB"
                "path"       = "/"
                "pool"       = "default"
                "size"       = "5GB"
            }
            type       = "disk"
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

johnlane:
[...]

PR #184, opened 2020-01-19
Fix Device Name Deletion

jtopjian:
This commit fixes a bug where the name attribute within a
device's properties was being deleted due to it being confused
with the device name itself.

For #171

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -450,7 +450,6 @@ func resourceLxdContainerRead(d *schema.ResourceData, meta interface{}) error {
 	for name, lxddevice := range container.Devices {
 		device := make(map[string]interface{})
 		device["name"] = name
-		delete(lxddevice, "name")
 		device["type"] = lxddevice["type"]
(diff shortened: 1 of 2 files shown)


jtopjian:
@sl1pm4t The test failures are for the "edge" channel. It's quite possible edge is broken right now. Is there any way we can merge this despite the test failures?

sl1pm4t:
@jtopjian - I've merged the change. Do you want me to cut a release? 

Also, do we need / want the edge channel to still be included in the builds?

jtopjian:
> Do you want me to cut a release?

Sure, I think we're due for a release. Thanks!

> Also, do we need / want the edge channel to still be included in the builds?

I've been thinking about that. I was hoping I'd get some time to dig into why edge is failing, but haven't been able to. Maybe we can just configure Travis to allow failures on edge?

sl1pm4t:
> Maybe we can just configure Travis to allow failures on edge?

I've added the edge builds to the `allow_faillures` list in the travis config.

jtopjian:
Thanks, Matt!

PR #194, opened 2020-05-13
Fix absence of linux.* attribute in config

### LINKED ISSUE #193: LXD provider does not work with "linux.kernel_modules" config property
When I use `linux.kernel_modules` in the `config` section of the `lxd_container`, provider passes it to LXD and properly creates the container, but after checking that the plan is equal to the applied state the property is absent in the state and as a result it produces changes.
As for "linux.kernel_modules" property it tries to re-create the container which is not valid behavior in this circumstances.

arren-ru:
Added support for linux.kernel_modules attribute in config section.

Fixes #193

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -405,6 +405,8 @@ func resourceLxdContainerRead(d *schema.ResourceData, meta interface{}) error {
 			config[k] = v
 		} else if strings.HasPrefix(k, "raw.") {
 			config[k] = v
+		} else if strings.HasPrefix(k, "linux.") {
+			config[k] = v
 		} else if strings.HasPrefix(k, "security.") {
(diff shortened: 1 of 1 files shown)


arren-ru:
Also I found a bunch of attributes which are not being merged into config schema.
https://lxd.readthedocs.io/en/latest/instances/#keyvalue-configuration
I would add all the namespaces listed on that page.

jtopjian:
@arren-ru Thanks for submitting this. Can you rebase this on top of master? That should fix the Travis issue.

arren-ru:
> @arren-ru Thanks for submitting this. Can you rebase this on top of master? That should fix the Travis issue.

Seems to be I already did that :)

jtopjian:
Ah, yes - my mistake.

PR #205, opened 2020-09-05
Updates required to publish to the Terraform Registry

jtopjian:


jtopjian:
@sl1pm4t I went through the process described here (https://www.terraform.io/docs/registry/providers/publishing.html) to publish my Sensu provider (https://registry.terraform.io/providers/jtopjian/sensu/latest). It was fairly straightforward.

The changes in this PR will take care of 80% of things. The other 20% are things you'll need to do:

1. Generate a GPG key
2. Sign in at registry.terraform.io
3. Add the GPG public key
4. Add the GPG private key and passphrase as secrets to this repo

arren-ru:
@jtopjian 
btw, I already published some versions here – https://registry.terraform.io/providers/arren-ru/lxd
Also I found that docs hierarchy is not correct for publishing at the registry.

PR #207, opened 2020-09-19
Migrate to the Terraform Plugin SDK

jgraichen:
While getting more comfortable around the provider and the terraform plugin environment I've decided to migrate the provider to the standalone plugin SDK as recommended here: https://www.terraform.io/docs/extend/plugin-sdk.html.

> » Why a separate module?
> 
> While the helper/* and other packages in Terraform Core has served us well, in order for provider development to evolve, the SDK needed to break out into its own repository. Terraform Core's versioning has been oriented towards practitioners. With the "unofficial" SDK existing in the core repository, the SDK becomes tied to Core releases and cannot follow semantic versioning. The new standalone SDK github.com/hashicorp/terraform-plugin-sdk follows sematic versioning starting with v1.0.0.

This might ease maintaining and appears to result in a bit smaller binaries too:

> -rwxr-xr-x  1 - -  47M Sep 19 16:41 terraform-provider-lxd
> -rwxr-xr-x  1 - -  49M Sep 19 16:42 terraform-provider-lxd-master


jtopjian (approved):
LGTM - thank you for doing this.

PR #209, opened 2020-10-02
Adds new resource lxd_publish_image

KLIM8D:
I have added a new resource that makes it possible to create a LXD image from a stopped container. Possible through lxc with `lxc publish`

### DIFF
--- a/docs/resources/README.md
+++ b/docs/resources/README.md
@@ -3,6 +3,7 @@
 ### Image
 
 * [`lxd_cached_image`](lxd_cached_image.md)
+* [`lxd_publish_image`](lxd_publish_image.md)
 
 ### Container
 
--- a/docs/resources/lxd_publish_image.md
+++ b/docs/resources/lxd_publish_image.md
@@ -1,0 +1,50 @@
+# lxd_publish_image
+
+Create a LXD image from a container
+
+## Example Usage
+
+```hcl
+resource "lxd_cached_image" "xenial" {
+  source_remote = "ubuntu"
+  source_image  = "xenial/amd64"
+}
+
+resource "lxd_container" "test1" {
+  name      = "test1"
+  image     = "${lxd_cached_image.xenial.fingerprint}"
+  profiles  = ["default"]
+
+  start_container = false
+}
+
+resource "lxd_publish_image" "test1" {
+  depends_on = [ lxd_container.test1 ]
+
+  container = "test1"
+  aliases   = [ "test1_img" ]
+}
+
+```
+
+## Argument Reference
+
+* `container` - *Required* - The name of the container
+
+* `aliases` - *Optional* - A list of aliases to assign to the image 
+
+* `properties` - *Optional* - A map of properties to assign to the image 
+
+* `public` - *Optional* - Whether the image can be downloaded by untrusted users.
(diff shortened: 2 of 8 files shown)


jtopjian:
@KLIM8D Thank you for submitting this.

Maybe I'm mistaken, but `lxd_publish_image` feels more like a task-based action rather than an idempotent resource. Can you explain how this resource is useful for managing the full lifecycle of a resource rather than being a one-off task?

Does the published image get updated each time `terraform apply` is run? How does the published image maintain a persistent state?

KLIM8D:
I find it kinda similar to the `lxd_snapshot` resource. It might be a task-based action as you mentioned, but let me explain how I use it.

I have a base image that installs and sets up a bunch of common stuff used by all my containers. So instead of doing this in every new container, I would like containers to launch from this image. This could of course be done manually with `lxc publish`, but I would like for this to be part of my Terraform configuration so I can track changes, tag (create aliases) images based on builds etc.

I've added a new triggers property, inspired by `null_resource`.

Below an example of how I use it to build this base image.


```
resource "lxd_cached_image" "ubuntu" {
  source_remote = "ubuntu"
  source_image = "20.04"
}

resource "lxd_container" "base_container" {
  name     = "testbaseimage"
  image    = lxd_cached_image.ubuntu.fingerprint
  profiles = ["default"]

  provisioner "local-exec" {
    command = "lxc exec local:${self.name} dhclient eth0"
  }

  provisioner "local-exec" {
    command = "lxc exec local:${self.name} apt-get update"
  }
}

resource "null_resource" "install" {
  depends_on = [ lxd_container.base_container ]

  triggers = {
    "has_changed" = local.build_ver
  }

  provisioner "local-exec" {
    command = "lxc start local:${lxd_container.base_container.name}  || true"
  }

  provisioner "local-exec" {
    command = "lxc exec local:${lxd_container.base_container.name} -- sh -c \"apt-get install -y golang-go\""
  }

  provisioner "local-exec" {
    command = "lxc exec local:${lxd_container.base_container.name} -- sh -c \"apt-get install -y gcc\""
  }

  provisioner "local-exec" {
    command = "lxc stop local:${lxd_container.base_container.name} || true"
  }
}

resource "lxd_publish_image" "base_image" {
  depends_on = [ null_resource.install ]

  container = lxd_container.base_container.name
  aliases = [ "base_image:${local.build_ver}" ]
  properties = {
    "name": "base_image"
    "os": "ubuntu"
    "version": "20.04"
    "build_ver": local.build_ver
  }

  triggers = {
    build_ver = local.build_ver
  }
}
```

jtopjian:
This looks to be a workflow more suitable for Packer. However,

> I find it kinda similar to the lxd_snapshot resource.

You're right - this resource is also rather debatable. But since it's already included, it's not fair to exclude this resource. 

I'm okay with merging this, though it looks like this branch needs to be rebased on top of master - would you be able to do that at your convenience?

KLIM8D:
Cool! I've added missing docs for triggers property and rebased on top of master.

jtopjian:
@KLIM8D This looks good to me. Was there anything else you wanted to do or is this ready to be merged?

KLIM8D:
It's ready :)

jtopjian (approved):
LGTM - thank you!

PR #212, opened 2020-10-22
Add target attribute to container resource for LXD cluster

jgraichen:
The target attributes allows placing a container on a specific node in a LXD cluster.

---

I have not found a useful way for testing without setting up a multi-node LXD cluster. Are you fine with skipping tests here?

### DIFF
--- a/docs/resources/lxd_container.md
+++ b/docs/resources/lxd_container.md
@@ -83,6 +83,8 @@ resource "lxd_container" "container1" {
 * `wait_for_network` - *Optional* - Boolean indicating if the provider should wait for the container's network address to become available during creation.
   Valid values are `true` and `false`. Defaults to `true`.
 
+* `target` - *Optional* - Specify a target node in a cluster.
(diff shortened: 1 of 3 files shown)


jtopjian:
@jgraichen Thanks so much for working on this.

For testing, would it be possible to include a draft acceptance test but just skip it? That will at least prepare for when we do have a testing environment that supports clustering.

I _think_ the following will work:

```go
func TestAccContainer_clusterTarget(t *testing.T) {
    t.Skip("Test environment does not support clustering yet")
    ...
}
```

But let me know if you're having difficulties writing a test in general.

jgraichen:
Will try that, thanks.

jgraichen:
Added a test case scheduling two containers on the same node. Without a target they should be spread across nodes. The test is skipped as proposed. Worked like a charm, thanks!

jtopjian (approved):
LGTM - thank you!

Issue #202, opened 2020-08-24
Support LXD 4.0 virtual machines


jgraichen:
LXD 4.0 introduced virtual machine support. A new "instances" API had been added to manage containers and virtual machines using a uniform interface. I'd like to have the ability to manage LXD virtual machines with this provider too.

A very small quick fix would be adding a type field to `lxc_container` resources and using the new API (draft available in [00c38a8](https://github.com/jgraichen/terraform-provider-lxd/commit/00c38a8f2284d70e7e601fdf68449a0a6d952c19)). It does look a bit funny to schedule `lxd_container` resources that are of type "virtual-machines", though.

Another idea would be to add a new resource mirror the unified API e.g. `lxd_instance`. It might be a chance to change other things in non-compatible way, e.g. the `config = {}` behavior.

I'd like to invest some time and contribute but would be interested to have this coordinated with the maintainers and how the provider should evolve.

jgraichen:
> Since there is a dedicated API for instances, this is the best way to go.

It is not a standalone API for virtual machines, it is a new API to manage containers & virtual machines (https://lxd.readthedocs.io/en/latest/rest-api/#instances-containers-and-virtual-machines). This API creates virtual machines just like containers based on an additional `type = container | virtual-machine` attribute. Both instance types support the same configuration options, etc.

Having a single resource for containers and virtual machines means the wished type can be specified as an attribute, e.g. like this:

```
resource "lxd_instance" "test" {
  name = "test"
  type = "virtual-machine"
}
```

This is quite nice because it is easy to switch between containers and virtual machines without having to change the resource type. It could very well replace "lxd_container", or we could, similar to the LXD API expose three resources, `lxd_container`, `lxd_virtual_machine` and `lxd_instance`, but I would not like that.

> ~ It might be a chance to change other things in non-compatible way, e.g. the config = {} behavior.
> Which issue is this?

I am not sure. There was a discussion somewhere if changes to `config` should force a recreation of a container or not. I myself had issues a few time with changed config values but not wanting to recreate the container. I applied the changes manually to have it in sync again.

jtopjian:
> It is not a standalone API for virtual machines, it is a new API to manage containers & virtual machines

Right, understood.

I haven't read the API in detail nor looked up the actual function calls in the LXD library, so I might be missing key information.

[...]

PR #215, opened 2020-12-05
Support LXD virtual machine types

jgraichen:
This MR adds a new attribute `type` to the `lxd_container` resources which is mapped to the LXD instance type, and can be used to create virtual machines instead of containers with recent LXD releases.

Unfortunately, I do not have enough time right now to create a new resource type as discussed in #202 but this MR provided a backward-compatible way to define virtual machines using the existing resource. The API should be compatible with much older LXD versions as the LXD client itself will fallback to the container only API when the server does not support the newer instances API (on a HTTP REST level).

With this a basic support for virtual machines would be ready.

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -33,6 +33,14 @@ func resourceLxdContainer() *schema.Resource {
 				Required: true,
 			},
 
+			"type": {
+				Type:         schema.TypeString,
+				ForceNew:     true,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: resourceLxdValidateInstanceType,
+			},
+
 			"remote": {
 				Type:     schema.TypeString,
 				ForceNew: true,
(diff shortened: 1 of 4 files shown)


lxd/resource_lxd_container.go
```diff
@@ -247,13 +255,24 @@ func resourceLxdContainerCreate(d *schema.ResourceData, meta interface{}) error
...
+	instanceType := d.Get("type").(string)
+	if instanceType == "container" {
+		createReq.Type = api.InstanceTypeContainer
+	} else if instanceType == "virtual-machine" {
+		createReq.Type = api.InstanceTypeVM
+	} else if instanceType == "" {
+		createReq.Type = api.InstanceTypeAny
+	} else {
+		return fmt.Errorf("invalid type: %s", instanceType)
+	}
```
jtopjian, line 274: I think an `else if` clause for `any` needs to be added?

jgraichen: Thanks for the hint. I forgot to remove `any` at all, because LXD will always change `any` to either `container` or `virtual-machine`. This confuses terraform as it will always result in new changed to apply. The default blank string behaves like `any`.

jgraichen:
I had to skip the virtual-machine test case as Travis CI seems to not support virtualization at all.

lxd/resource_lxd_container.go
jtopjian: Do you think having a blank option is intuitive to users? Meaning, if a user chooses a blank string (or not set the option at all), the outcome will be what the user expects?

jgraichen: A blank string as the attribute value is invalid as it is rejected by `resourceLxdValidateInstanceType`:

```
Error: Instance must have a type of: [container virtual-machine]

  on main.tf line 15, in resource "lxd_container" "testtest":
  15: resource "lxd_container" "testtest" {
            type = ""
```

If basically means, if `type` is given, use it, otherwise give me any kind of instance. I am not sure if LXD has a setting for a default instance type, but that would be supported here without change.

jgraichen:
Rebased PR and tests are passing. Ready to be merged from my side.

@jtopjian @sl1pm4t Thanks for all your work today! It is a pleasure.

jtopjian (approved):
LGTM - thank you!

PR #216, opened 2020-12-05
Update alpine image used in tests

jgraichen:
The alpine/3.9 images has reached EOL and is not longer distributed by the LXD image servers.

---

I did thought about changing this to `alpine/edge` which should always be available but I am not sure about implications or previous experience and only changed to `alpine/3.12`.

With this the test suite is passing again. I'll rebase #215 after this has been merged.


jtopjian (approved):
LGTM - thank you!

PR #227, opened 2020-12-15
Don't Cause Rebuilds on Config Changes

jtopjian:
This commit removes the logic that causes containers
to be rebuilt on config changes.


jgraichen:
Can we get a new release that does include this change? 😇

PR #240, opened 2021-10-05
Do update container config on change

jgraichen:
The `config` attribute doesn't force a new resource anymore, but is not updated at all. Changes simply were ignored.

This PR adds the missing parts to update config keys when updating a container resource.


jgraichen:
This is the only failure I could identify:

```
=== RUN   TestAccContainer_basicEphemeral
    testing.go:745: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.
        
        Error: errors during apply: Error waiting for container (topical-pigeon) to stop: Not Found
        
        State: lxd_container.container1:
          ID = topical-pigeon
          provider = provider.lxd
          config.% = 0
          device.# = 0
          ephemeral = true
          file.# = 0
          image = images:alpine/3.12/amd64
          ip_address = 10.242.212.45
          ipv4_address = 10.242.212.45
          ipv6_address = fd42:196e:2bcb:4364:216:3eff:feee:29fc
          limits.% = 0
          mac_address = 00:16:3e:ee:29:fc
          name = topical-pigeon
          privileged = false
          profiles.# = 1
          profiles.0 = default
          start_container = true
          status = Running
          target = none
          type = container
          wait_for_network = true
--- FAIL: TestAccContainer_basicEphemeral (22.73s)
```

jgraichen:
I've rebased the PR.

jgraichen:
As far as I see it, it is indeed ready.

I had some container where I needed to update the config without a rebuild, but that isn't available in `1.5.0`. I used a local build of the provider, and... nothing happened. This MR fixed that and worked.

I'd love to see a `1.6.0` release with this included. Configuring a local provider binary is quite a hassle in newer terraform versions. Thanks a lot!

PR #252, opened 2022-01-14
Expose attribute location for created volumes

lastkrick:
This is very useful in cluster mode.
If you will create volume without providing target it could be created on any node in the cluster, and you must to create container on the same target to be able use this volume.


jtopjian:
@lastkrick This looks good to me. Let me know if you're ready to have this merged.

lastkrick:
@jtopjian Thank you! Let's merge it

jtopjian:
@lastkrick Thanks for working on this. There should be a new release shortly with this included.

PR #259, opened 2022-05-04
feat: keep device name

ComputerBaer:
This is required for configurations like this to keep the interface names in the containers:
```
  device {
    name = "eth0"
    type = "nic"

    properties = {
      name    = "eth0"
      nictype = "bridged"
      parent  = "network1"
    }
  }

  device {
    name = "eth1"
    type = "nic"

    properties = {
      name    = "eth1"
      nictype = "bridged"
      parent  = "network2"
    }
  }
```

Without this change, the provider always want to apply changes:
```
      + device {
          + name       = "eth0"
          + properties = {
              + "name"    = "eth0"
              + "nictype" = "bridged"
              + "parent"  = "network1"
            }
          + type       = "nic"
        }
      - device {
          - name       = "eth0" -> null
          - properties = {
              - "nictype" = "bridged"
              - "parent"  = "network1"
            } -> null
          - type       = "nic" -> null
        }
      + device {
          + name       = "eth1"
          + properties = {
              + "name"    = "eth1"
              + "nictype" = "bridged"
              + "parent"  = "network2"
            }
          + type       = "nic"
        }
      - device {
          - name       = "eth1" -> null
          - properties = {
              - "nictype" = "bridged"
              - "parent"  = "network2"
            } -> null
          - type       = "nic" -> null
        }
```

### DIFF
--- a/lxd/resource_lxd_container_test.go
+++ b/lxd/resource_lxd_container_test.go
@@ -464,6 +464,13 @@ func TestAccContainer_withDevice(t *testing.T) {
 	var container api.Container
 	containerName := strings.ToLower(petname.Generate(2, "-"))
 
+	device := map[string]string{
+		"type":    "nic",
+		"name":    "bar",
+		"nictype": "bridged",
+		"parent":  "lxdbr0",
+	}
+
 	resource.Test(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
 		Providers: testAccProviders,
@@ -473,6 +480,7 @@ func TestAccContainer_withDevice(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testAccContainerRunning(t, "lxd_container.container1", &container),
 					resource.TestCheckResourceAttr("lxd_container.container1", "name", containerName),
+					testAccContainerDevice(&container, "foo", device),
 				),
 			},
 		},
--- a/lxd/resource_lxd_profile.go
+++ b/lxd/resource_lxd_profile.go
@@ -119,7 +119,6 @@ func resourceLxdProfileRead(d *schema.ResourceData, meta interface{}) error {
 	for name, lxddevice := range profile.Devices {
 		device := make(map[string]interface{})
 		device["name"] = name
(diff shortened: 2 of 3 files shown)


jtopjian:
@ComputerBaer Thanks for submitting this. 

I'm confused about why the tests are failing - I'm unable to reproduce the test failure myself.

However, I'm curious about why this change is needed. We have a lot of tests that define devices and Terraform isn't reporting any changes on them. One example is here: 

https://github.com/terraform-lxd/terraform-provider-lxd/blob/master/lxd/resource_lxd_container_test.go#L1046-L1081

Given the issue you're reporting, I would have expected that test to fail.

Are you able to create an acceptance test that reproduces this problem? That might help us narrow down the root cause.

ComputerBaer:
Oh, yes, I missed a detail in the description. I defined the devices in a profile and not in a container.

I will try to create a test tomorrow.

jtopjian:
> I defined the devices in a profile and not in a container.

Yes, you're right. That should have clicked when I saw the file you edited 🙂 

If you could create a test with this PR, that would help. If not, I can add one when I get some time.

ComputerBaer:
The container removes only type from the device map and not the name:
 https://github.com/terraform-lxd/terraform-provider-lxd/blob/master/lxd/resource_lxd_container.go#L526-L531

That should be the reason why the container tests with device names do not fail.

jtopjian:
You're right. This change makes sense, then. If we can get a test in place to show this is working and to help prevent regression, we should be good to go.

ComputerBaer:
Could it be that the test don't check if the device is present or any of its properties?
https://github.com/terraform-lxd/terraform-provider-lxd/blob/master/lxd/resource_lxd_container_test.go#L463-L480

(These tests will be my first lines in Go)

jtopjian:
The test framework will automatically do an additional "terraform apply" and make sure that no changes happen. This should be good enough to make sure that the issue described here is fixed.

jtopjian:
The test failures have been fixed in #261.

This looks good to me. Thank you for both reporting the problem and submitting a fix!

Issue #260, opened 2022-05-05
Provide valid image in provider docs


dontlaugh:
A fully-qualified image name isn't provided in the docs right now. This string "images:ubuntu/22.04" works, but just "ubuntu" doesn't.

```hcl

resource "lxd_container" "test1" {
  name      = "test1"
  image     = "images:ubuntu/22.04"
  ephemeral = false

  config = {
    "boot.autostart" = true
  }

  limits = {
    cpu = 2
  }
}

```

jtopjian:
Fixed in #265

PR #265, opened 2022-08-28
Specify valid image in container resource example

dontlaugh:
The public "images:" server prefix is a good choice. Ubuntu 22.04 is an LTS and should be available for a long time.


dontlaugh:
Fixes #260

jtopjian (approved):
The test suite is failing, but that's unrelated to this PR. This LGTM, thank you!

PR #271, opened 2022-10-27
Add new device types to the docs

al20ov:
I added a little bit of documentation as well as an example of how to use proxy devices in order to forward ports from the host to a container.


jtopjian (approved):
This looks good to me, thank you!

Issue #272, opened 2022-10-31
Segfault when using `lxd_publish_image` for a `virtual-machine` instance


Unitiser:
# Segfault when using `lxd_publish_image` for a `virtual-machine` instance

The `terraform-lxd/lxd` plugin segfaults when we try to publish an image for a `virtual-machine` instance.

When trying to understand the issue, I noticed that :
- The plugin crashes before the build begins
- Publishing regular container work as expected

## Sample configuration

Here is a sample configuration to reproduce the issue :

```
terraform {
  required_providers {
    lxd = {
      source = "terraform-lxd/lxd"
    }
  }
}

provider "lxd" {
  lxd_remote {
    name     = "${var.lxd_provider_name}"
    scheme   = "${var.lxd_provider_scheme}"
    address  = "${var.lxd_provider_address}"
    default  = true
    port     = "${var.lxd_provider_port}"
  }
}

resource "lxd_network" "lab" {
  name = "lab2"

  config = {
    "ipv4.address" = "10.0.2.1/24"
    "ipv4.nat"     = "true"
    "ipv6.address" = "none"
  }
}

resource "lxd_profile" "lab" {
  name = "lab2"

  device {
    name = "eth0"
    type = "nic"

    properties = {
      nictype = "bridged"
      parent  = "${lxd_network.lab.name}"
    }
  }
  
  device {
    name = "root"
    type = "disk"
    
    properties = {
      pool = "btrfs-pool"
      path = "/"
    }
  }
}

resource "lxd_container" "test" {
  name      = "test1"
  image     = "win10"
  type      = "virtual-machine"
  
  ephemeral = false
  profiles  = ["${lxd_profile.lab.name}"]
  
  config = {
    "boot.autostart" = true
    "security.secureboot" = false
    "limits.cpu" = "2"
    "limits.memory" = "4GB" 
  }
}

resource "lxd_publish_image" "test" {
  container = "test1"
  aliases   = [ "test1" ]
}
```

## Stacktrace

```
Stack trace from the terraform-provider-lxd_v1.7.2 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xdb1269]

goroutine 44 [running]:
[...]

PR #276, opened 2022-11-01
Replace 'GetContainerState' with 'GetInstanceState' to support VM imaging

Unitiser:
Fix for `lxd_publish` following the troubleshooting of https://github.com/terraform-lxd/terraform-provider-lxd/issues/272.


Unitiser:
@jtopjian I have not done extensive tests, but the terraform script cited in the issue works with the modified code.

That being said, the function is only used to check if the state of the vm/container is running or not and the state object is identical. Therefore I'm quite confident this wont have any side effect.

jtopjian (approved):
LGTM - thank you!

Issue #278, opened 2022-12-14
Upgrade LXD API dependency


dorkamotorka:
This version https://github.com/terraform-lxd/terraform-provider-lxd/blob/d6bb607fd8083d5e4654141e930f659a9feffc30/go.mod#L15 lacks things like **config/limits.intances** and **config/features.networks** which is essential in case we want **project** configuration parameters to make sense in the **lxd_network**

PR #279, opened 2022-12-14
WIP: Add lxd_project resource type

dorkamotorka:
**project** configuration parameter should work with every other resource (there is just a slight "issue" with **lxd_network**. Look here for an update on the issue: #278)  

I also added basic unit-testing and verified the functionality on a local setup. This PR still lacks documentation, but I plan to do this once all the software-related things are successfully done.

@jtopjian Would you mind checking out the code, and perhaps approving running the workflow so we can give the tests a whirl? Thanks!


jtopjian:
@dorkamotorka To get around the test suite requiring approval, submit a small PR with an update/change to something simple like documentation. I'll merge it and that should help.

dorkamotorka:
@jtopjian the test suite seems to run fine and the *_project tests seem to work fine until the actions just fail. Is it possible that this is due to a time limit or something because there seem to be NO tests that actually fail?

jtopjian:
@dorkamotorka It looks like the following test is failing:

```
=== RUN   TestAccNetwork_project
    testing.go:705: Step 0 error: Check failed: Check 2/2 error: Not found: lxd_network.eth0
--- FAIL: TestAccNetwork_project (4.45s)
```

dorkamotorka:
@jtopjian Sorry, I missed that - thanks for your patience. The test suite now passes which was expected, but I guess the real question is whether the tests I created are sufficient.

jtopjian:
@dorkamotorka Nice work :)

I skimmed the tests and they look comparable to what we already have, so that should be good. The tests in this project rely heavily on the resource being updated via the API and then triggering a mismatch if one exists. The key piece of code is this area:

https://github.com/terraform-lxd/terraform-provider-lxd/pull/279/files#diff-dbeb3335f87880a6b9890154c58c29425b3d552c639f7fd3d76ef349cd4acf6dR91-R100

(which it looks like there might be a typo here, setting name to config)

The `d.Set()` stuff sets the resource property to what was retrieved by the API. So if, for some reason, the resource created a project called "Foo" but the API returned "Bar", Terraform would flag it due to this bit of code (including in tests). Without this bit of code, the mismatch would continue unnoticed.

Let me know if that makes sense.

dorkamotorka:
Nice catch of the typo! If I understood you correctly the issue should be fixed.

lxd/resource_lxd_cached_image.go
```diff
@@ -88,6 +94,10 @@ func resourceLxdCachedImageCreate(d *schema.ResourceData, meta interface{}) erro
 	if err != nil {
 		return err
 	}
+	if p, ok := d.GetOk("project"); ok && p != "" {
```
jtopjian, line 97: nit: I recommend using `v, ok` which is more consistent with other coding patterns in this project. This comment applies to all occurrences of where you have:

```
if p, ok := ...
```

jtopjian (changes requested):
@dorkamotorka Thank you for working on this. In general, this is really good. I just have a few requests about documentation wording and coding consistency. Please let me know if you have any questions.

dorkamotorka:
Thanks for the review 💯  I think I managed to apply all your suggestions

jtopjian:
@dorkamotorka Thanks. This all looks good to me. Were you planning to do any other work or is this ready to be merged?

dorkamotorka:
We can merge this. Thanks for the review 💯

dontlaugh:
Thank you @dorkamotorka

PR #289, opened 2023-04-06
Terraform Plugin SDK v2 Upgrade

adamcstephens:
I ran the upgrade migration tool, and then addressed all visible errors. 

https://developer.hashicorp.com/terraform/plugin/sdkv2/guides/v2-upgrade-guide
https://github.com/hashicorp/tf-sdk-migrator#tf-sdk-migrator-v2upgrade-migrate-from-sdkv1-to-sdkv2

This is pretty similar to #243, without some of the other changes. Also marking #285 closed with this, since it takes x/net to 0.8.0

Closes #243 
Closes #285


adamcstephens:
Thanks for approving the tests. I'll see if I can diagnose what's going wrong, but they run successfully locally. :/

jtopjian:
@adamcstephens I think you can now amend the PRs and the tests will run automatically. If that doesn't work, try opening a new PR. 

Let me know if you have any questions or need anything investigated. It might take me a day or two, but I'll get to it 🙂

adamcstephens:
Hi @jtopjian, thanks for the ping. I've actually been working on this, but running into some issues with the tests. I also am familiarizing myself with the codebase, along with the terraform provider docs, given that I've never worked on one before. I did learn over the weekend that even sdk v2 is no longer even recommended and they are pushing people towards the Terraform Plugin Framework now. 

One of the things I'm struggling with in the tests is our provider testing. Specifically, we define a provider block in our configurations, such as https://github.com/terraform-lxd/terraform-provider-lxd/blob/0c6b4ebbc2d80fffcd6101ce153aaef0296db059/lxd/provider_test.go#L237-L238

This causes a duplicate provider conflict as the acceptance testing framework appears to bring its own provider initialization. I'm still exploring solutions, but I'm wondering if these tests are actually valuable? I also wonder if we should even support the multi-remote functionality at all, given that Terraform has native support for multiple provider blocks.

jtopjian:
> This causes a duplicate provider conflict as the acceptance testing framework appears to bring its own provider initialization. I'm still exploring solutions, but I'm wondering if these tests are actually valuable?

They were valuable at one time. If you want to try removing these tests to see if the other tests pass, it's worth a try. But there's a lot of logic in the provider configuration for connecting to an LXD remote, so I would caution about modifying anything related to that.

>  I also wonder if we should even support the multi-remote functionality at all, given that Terraform has native support for multiple provider blocks.

This support was added before LXD supported clustering and before Terraform allowed multiple provider declarations of the same provider, so it's a bit of a historical feature. I would say that removing this functionality would require a major version increment, which is possible to do, but my advice would be to try to upgrade all of the internal libraries first and then start making front-facing architecture changes like this.

adamcstephens:
> I would say that removing this functionality would require a major version increment, which is possible to do, but my advice would be to try to upgrade all of the internal libraries first and then start making front-facing architecture changes like this.

Yes, I 100% agree with this. I wouldn't want to make a change to this functionality without a clear deprecation and a major version change. For now I'm fully focused on getting the tests to pass and getting the dependencies up to date.

Most of the test issues are small, and some are actually fixed in #243 in what I thought were unrelated changes.

The one I've been unable to track down yet is a failure in the publish image tests, which is causing a panic.

```
=== RUN   TestAccPublishImage_aliases                                                                                                                         
panic: Invalid address to set: []string{"architecture"}                                                                                                       
```

https://github.com/adamcstephens/terraform-provider-lxd/blob/v2upgrade/lxd/resource_lxd_publish_image.go#L242-L244

jtopjian:
> The one I've been unable to track down yet is a failure in the publish image tests, which is causing a panic.

I'm quite certain that is because `fingerprint`, `architecture`, and `created_at` aren't defined in that resource's schema at all. Terraform must have been silently disregarding that in the past.

adamcstephens:
All acceptance tests are now passing. Let me know if you have any feedback, want me to squash commits, etc.

lxd/provider.go
```diff
@@ -125,37 +124,6 @@ func Provider() terraform.ResourceProvider {
...
-				Removed:  "Use `lxd_remote.name` instead.",
-			},
-
-			"remote_password": {
-				Type:      schema.TypeString,
-				Sensitive: true,
-				Optional:  true,
-				Removed:   "Use `lxd_remote.password` instead.",
-			},
-
```
jtopjian, line 158: Why are these being removed? If everything is still functional with them included, we should keep them and defer removal for a major update increment.

lxd/resource_lxd_container.go
```diff
@@ -160,6 +160,7 @@ func resourceLxdContainer() *schema.Resource {
 						"mode": {
 							Type:     schema.TypeString,
 							Optional: true,
+							Default:  "0755",
```
jtopjian, line 163: Is this needed for the SDK update? If not, let's remove it. Plus, 0755 could be insecure for a default mode.

lxd/resource_lxd_container.go
```diff
@@ -328,17 +329,6 @@ func resourceLxdContainerCreate(d *schema.ResourceData, meta interface{}) error
...
-	d.SetPartial("name")
-	d.SetPartial("image")
-	d.SetPartial("profiles")
-	d.SetPartial("project")
-	d.SetPartial("ephemeral")
-	d.SetPartial("privileged")
-	d.SetPartial("config")
-	d.SetPartial("limits")
-	d.SetPartial("device")
-	d.SetPartial("remote")
```
jtopjian, line 340: Does `SetPartial` no longer work with the SDK update? If it still works, let's keep this and remove it in another PR (in trying to keep PRs scoped to a specific focus).

lxd/resource_lxd_container_file.go
```diff
@@ -45,18 +45,21 @@ func resourceLxdContainerFile() *schema.Resource {
...
 				ForceNew: true,
 				Optional: true,
+				Default:  0,
 			},
 
 			"mode": {
 				Type:     schema.TypeString,
 				ForceNew: true,
 				Optional: true,
+				Default:  "0755",
```
jtopjian, line 62: Same comment here as the other one regarding `mode`.

lxd/resource_lxd_publish_image.go
```diff
@@ -24,6 +24,21 @@ func resourceLxdPublishImage() *schema.Resource {
...
+			"architecture": {
+				Type:     schema.TypeString,
+				ForceNew: false,
+				Computed: true,
+			},
+			"created_at": {
+				Type:     schema.TypeInt,
+				ForceNew: false,
+				Computed: true,
+			},
```
jtopjian, line 41: Thank you for adding these. Please make sure they're included in the documentation, too.

jtopjian (changes requested):
@adamcstephens Thanks so much for your work on this. In general, this all looks really good and definitely a long-needed update.

I do have a few comments. Please let me know if you have any questions or would like to discuss them.

lxd/resource_lxd_publish_image.go
adamcstephens: Documentation updated.

adamcstephens (commented):
Updated a couple things and responded to questions/feedback.

lxd/resource_lxd_container.go
jtopjian: > I could probably fix the panic without moving the default if you'd prefer, but it seemed better practice to have the default be in the schema and not the code.

Yeah, that makes sense.

> Furthermore, this file mode is used for any directory creations as well, so 0755 could be an appropriate default for directories.

I'm pretty sure this was reviewed in the past because I would have questioned 0755 whenever it comes up. It's okay for now, but I do recommend looking at 0750 in the future. The issue isn't so much about directories, but files where anyone can see the contents of potentially sensitive files that are copied to the container. The end responsibility is on the user, though - they can always change the mode from the default.

jtopjian (approved):
This looks good to me. Really nice work - thank you very much!

lxd/resource_lxd_container.go
adamcstephens: I'd be happy to revisit this later. We should probably split the two modes and have two attributes, one for the file and one for any directories that need to be created. Then we can have a more restrictive default for the file itself.

jtopjian:
@adamcstephens Thank you again. Based on this work, I think you've done a great job. If you're interested in acting as a maintainer, I can set you up with the appropriate privileges in this repository. You can reach out to ask for help, history, advice, etc but also be able to take an active role here. Let me know.

adamcstephens:
@jtopjian I am definitely still interested in helping as a maintainer.

PR #292, opened 2023-04-19
migrate from sdk2 testing to terraform-plugin-testing framework

adamcstephens:
https://developer.hashicorp.com/terraform/plugin/testing/migrating

This is the current recommended testing framework. If and when an effort is undertaken to migrate the newer Plugin Framework, this will be a requirement. Either way, this was a relatively simple migration.


jtopjian:
@adamcstephens feel free to commit when you're comfortable. If you'd like me to review something, either ping me in a comment or use the "request review" feature.

PR #293, opened 2023-05-24
Add LXD volume copy resource

lukedirtwalker:
LXD volume copy is almost the same as an LXD volume, but for the creation we copy from an existing volume.


adamcstephens (changes requested):
I haven't had a change to fully review this myself, but please include documentation updates with resource changes.

adamcstephens:
Other than the missing docs, this looks good to me.

adamcstephens (approved):

adamcstephens (approved):
LGTM. Thank you! :bee:

PR #294, opened 2023-06-02
add content_type option to volume

### LINKED ISSUE #277: Unable to create a block custom volume
LXD supports 2 types of ["custom" storage volumes](https://linuxcontainers.org/lxd/docs/master/howto/storage_volumes/#create-the-volume): `filesystem` and `block`.

When creating a volume using the [lxd_volume](https://registry.terraform.io/providers/terraform-lxd/lxd/latest/docs/resources/volume) resource of the provider, we can specify the `type = "custom"` but we cannot choose the volume `content_type` which defaults to `filesystem`.

We should be able to specify the `content_type` to either `filesystem` or `block`.

### LINKED ISSUE #288: How to create storage volume with content_type block with terraform-lxd ?
I want to create a storage volume with content_type block like this one :

```yaml
config:
  size: 730GB
description: ""
name: ulxd-k0s-n1-zfspv-pool
type: custom
used_by:
- /1.0/instances/ulxd-k0s-n1
location: none
content_type: filesystem
project: default
created_at: 0001-01-01T00:00:00Z
```

I used this config :

```
additional_disks = {
             zfspv-pool = {
                pool = "nvme2n1",
                size = "730GB",
                content_type = "block"
            }  
        }

resource "lxd_volume" "volume" {
  for_each = var.additional_disks
  remote = var.remote == "" ? "${var.zone}-lxd-hypervisor-01": var.remote 
  name =  "${var.name}-${each.key}"
  pool = contains(keys(each.value),"pool") ? each.value["pool"] : var.storage_pool
  config = {
    size = contains(keys(each.value),"size") ? each.value["size"] : null
    content_type = contains(keys(each.value),"content_type") ? each.value["content_type"] : null
  }
}
```
but I get error that said there is no option ``content_type``,  and seems that it uses filesystem content type by default.

gboutry:
Fix #288
Fix #277

Being able to change content_type allows for the creation of block volumes.

### DIFF
--- a/docs/resources/volume.md
+++ b/docs/resources/volume.md
@@ -39,6 +39,8 @@ resource "lxd_volume" "volume1" {
 
 * `project` - *Optional* - Name of the project where the volume will be stored.
 
+* `content_type` - *Optional* - Volume content type (filesystem or block)
+
 ## Attribute Reference
 
 The following attributes are exported:
--- a/lxd/resource_lxd_volume.go
+++ b/lxd/resource_lxd_volume.go
@@ -69,6 +69,13 @@ func resourceLxdVolume() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 			},
+
+			"content_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Default:  "filesystem",
+			},
 		},
 	}
 }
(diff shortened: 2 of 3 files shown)


adamcstephens (approved):
Thanks! :bee: 

I couldn't find "filesystem" documented as the default for content_type, but it is in practice.

gboutry:
> Thanks! bee This PR is a drop-in for my own branch with this capability.
> 
> I couldn't find "filesystem" documented as the default for content_type, but it is in practice.

I got the default from here https://linuxcontainers.org/lxd/docs/latest/howto/storage_volumes/#create-the-volume

PR #306, opened 2023-07-10
Rename container resource to instance

### LINKED ISSUE #302: Rename container resource to instance
The current `lxd_container` resource in terraform-provider-lxd represents an LXD instance rather than an actual container. This naming convention can be a bit confusing since the container can not be of type virtual machine.

To address this confusion, I propose renaming the `lxd_container` resource to `lxd_instance`. This change will better reflect its purpose and align with LXD terminology. To ensure a smooth transition and avoid breaking changes, we can deprecate the `container` resource and keep it available for a certain period of time.

I can make a PR if the maintainers agree?

MusicDin:
This PR introduces a new resource `lxd_instance` which will replace an existing resource `lxd_container`.

Changes:
- Copied `lxd_container` to `lxd_instance` and updated functions, variables, and comments. In addition, locally used functions are moved to `lxd_instance`.
- Copied `lxd_container_file` to `lxd_instance_file` and updated functions, variables, and comments.
- Added depreciation messages to `lxd_container` and `lxd_container_file`
- Added new resources to the `provider`
- Added new resources to the docs and fixed some URLs

Fixes #302

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -25,6 +25,7 @@ func resourceLxdContainer() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: resourceLxdContainerImport,
 		},
+		DeprecationMessage: "lxd_container has been deprecated and will be removed. Please use lxd_instance instead.",
 
 		Schema: map[string]*schema.Schema{
 			"name": {
@@ -335,7 +336,7 @@ func resourceLxdContainerCreate(d *schema.ResourceData, meta interface{}) error 
 		for _, v := range files.([]interface{}) {
 			f := v.(map[string]interface{})
 			file := File{
-				ContainerName:     name,
+				InstanceName:      name,
 				TargetFile:        f["target_file"].(string),
 				Content:           f["content"].(string),
 				Source:            f["source"].(string),
@@ -345,7 +346,7 @@ func resourceLxdContainerCreate(d *schema.ResourceData, meta interface{}) error 
 				CreateDirectories: f["create_directories"].(bool),
 			}
 
-			if err := containerUploadFile(server, name, file); err != nil {
+			if err := instanceUploadFile(server, name, file); err != nil {
 				return err
 			}
 		}
@@ -663,7 +664,7 @@ func resourceLxdContainerUpdate(d *schema.ResourceData, meta interface{}) error 
 			f := v.(map[string]interface{})
 			targetFile := f["target_file"].(string)
 
-			if err := containerDeleteFile(server, name, targetFile); err != nil {
+			if err := instanceDeleteFile(server, name, targetFile); err != nil {
 				return err
 			}
 		}
@@ -671,7 +672,7 @@ func resourceLxdContainerUpdate(d *schema.ResourceData, meta interface{}) error 
 		for _, v := range newFiles.([]interface{}) {
 			f := v.(map[string]interface{})
 			newFile := File{
-				ContainerName:     name,
+				InstanceName:      name,
 				TargetFile:        f["target_file"].(string),
 				Content:           f["content"].(string),
 				Source:            f["source"].(string),
@@ -681,7 +682,7 @@ func resourceLxdContainerUpdate(d *schema.ResourceData, meta interface{}) error 
 				CreateDirectories: f["create_directories"].(bool),
 			}
 
-			if err := containerUploadFile(server, name, newFile); err != nil {
+			if err := instanceUploadFile(server, name, newFile); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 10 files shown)


jtopjian:
@MusicDin I've taken a quick look and this looks pretty good. I think there is a way to mark the whole resource as deprecated, though, instead of the individual attributes:

https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-removal

It should be possible to set `DeprecationMessage` in the schema.

docs/resources/container.md
adamcstephens, line 1: Can you please add a deprecation warning to the top of docs for `lxd_container` and `lxd_container_file` pointing people to the new instance resources?

lxd/resource_lxd_container.go
```diff
@@ -859,38 +860,3 @@ func resourceLxdContainerWaitForNetwork(server lxd.ContainerServer, name string)
 		return st, "NOT FOUND", nil
 	}
 }
-
-// Suppress Diff on empty name
-func suppressImageDifferences(k, old, new string, d *schema.ResourceData) bool {
```
adamcstephens, line 864: Would it make sense to put these in `shared.go` instead?

adamcstephens (changes requested):
A couple small items but otherwise this looks good. Deprecation message and importing new resource name works as I'd expect.

Example deprecation


```
│ Warning: Deprecated Resource
│ 
│   with lxd_container.ci,
│   on config.tf.json line 49, in resource.lxd_container.ci:
│   49:       },
│ 
│ lxd_container has been deprecated and will be removed. Please use lxd_instance instead.
│ 
│ (and 25 more similar warnings elsewhere)
```

lxd/resource_lxd_container.go
MusicDin: We can put in in `shared.go`, or maybe have something like `instance_utils.go`, as they are currently instance specific.

What are your thoughts on that?

docs/resources/container.md
MusicDin: Absolutely. I have totally missed that one.

lxd/resource_lxd_container.go
adamcstephens: Actually, now that I've thought about this some more I don't mind them located with the instance definition. So feel free to leave them as this PR has, or `instance_utils.go` sounds good to. Your decision.

lxd/resource_lxd_instance.go
```diff
@@ -0,0 +1,896 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"mac_address": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"start_instance": {
```
adamcstephens, line 202: Since we're renaming this anyway, please rename this to `start_on_create` to clarify that it is only use to control the creation state. Also, if you don't mind, add it to the documentation. If not, I can add it after merge.

lxd/resource_lxd_container.go
MusicDin: Agree, we can leave it as it is for now as there are only a few functions, and separate the logic later if needed.

lxd/resource_lxd_instance.go
MusicDin: Done.

docs/resources/container_file.md
```diff
@@ -1,5 +1,7 @@
 # lxd_container_file
 
+!> Resource `lxd_container_file` has been deprecated and will be removed. Please use [`lxd_instance_file`](../instance_file.md) instead.
```
MusicDin, line 3: I've added a warning with the link to the new resource `... Please use [lxd_instance_file](../instance_file.md) instead.`. 

Can you confirm the url pointing to a new resource is correct, to ensure there is no broken links?

adamcstephens: I can't find any guidance on links between resources for terraform docs. Note that these get generated into what shows up on the registry at https://registry.terraform.io/providers/terraform-lxd/lxd/latest/docs/resources/container . Given that, I suspect the link is not correct since it has `.md`, but without deploying a release I know no other way to test them. I think it should be sufficient to just mention the new resource by name and let the users navigate using the sidebar.

jtopjian: I agree here. 

One option would be to dive into some other Terraform provider repos and try to find an example of inter-linking within documentation. Short of that, though, just mentioning the resource name is the best way to ensure there isn't a broken link.

adamcstephens: I looked at a few but they either didn’t have any links or were using the `website/docs` structure which I didn’t explore.

MusicDin: I had the same issue with finding any official guidance on links between resources.
I've removed the links, leaving just the warnings.

adamcstephens:
This looks good to me. I'm going to go ahead and merge this in an effort to not make this PR any larger. We'll need to do some more cleanup before it'll be in a release. There are still a number of tests and documentation examples that reference `lxd_container`.

adamcstephens:
Thanks for your contribution @MusicDin!

PR #312, opened 2023-07-13
volume: update content_type on read

### LINKED ISSUE #310: upgrade to 1.10 fails due to introduction of content_type
I just did a painless upgrade from 1.7.3 to 1.9.1 (through all intermediate releases) without a glitch.

However, it seems that the introduction of content_type = "filesystem" in #294 forces me to destroy and recreate volumes, causing quite a bit of downtime, with snapshot/clone tricks galore.

Is there a way to make this enhancement non-destructive? it seems that the implicit default was "filesystem" in versions 1.9.1 and lower. 
One of my preferred workaround strategies would be to have "terraform import" to the rescue, but that's not available for lxd_volume unfortunately either.

adamcstephens:
Fixes #310

### DIFF
--- a/lxd/resource_lxd_volume.go
+++ b/lxd/resource_lxd_volume.go
@@ -155,6 +155,7 @@ func resourceLxdVolumeRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("config", newConfig)
 	d.Set("expanded_config", volume.Config)
 	d.Set("location", volume.Location)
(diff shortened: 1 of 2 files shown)


adamcstephens:
@jtopjian this works in my testing. Can you take a look to verify this matches what you were thinking? Thanks.

jtopjian (approved):
@adamcstephens Yes, this is exactly what I was thinking - thank you!

PR #314, opened 2023-07-13
Update actions, go modules, and enable dependabot

adamcstephens:
I noticed some actions warnings, and had been exploring dependabot already. This won't update the lxd client unfortunately, but I'll need to investigate that another time.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,28 +20,28 @@ jobs:
     steps:
       -
         name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       -
         name: Unshallow
         run: git fetch --prune --unshallow
       -
         name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v4
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -20,28 +20,28 @@ jobs:
...
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v4
         with:
-          go-version: "1.20"
+          go-version: '1.20'
       -
         name: Import GPG key
         id: import_gpg
-        uses: paultyng/ghaction-import-gpg@v2.1.0
+        uses: crazy-max/ghaction-import-gpg@v5
```
adamcstephens, line 35: Hashicorp had a fork of this, but have deprecated it and are sending people to crazy-max now. https://github.com/hashicorp/ghaction-import-gpg#warning-this-action-has-been-deprecated

adamcstephens:
@jtopjian let me know if you have any thoughts on this. i obviously won't be able to test the release workflow changes until the next release is cut.

PR #318, opened 2023-07-23
Fix file upload on virtual machine creation

### LINKED ISSUE #315: Cannot use "file {" blocks on "lxd_container" of type "virtual-machine"
I'm trying to use `file` to install a script on a new virtual machine:

```terraform
resource "lxd_container" "v1" {
  name = "v1"
  type = "virtual-machine"
  image = "images:ubuntu/22.04"
  profiles = ["default"]

  file {
    source = "scripts/docker-setup.sh"
    target_file = "/scripts/docker-setup.sh"
    uid = 0
    gid = 0
    mode = "700"
    create_directories = true
  }
}
```

When I run `terraform apply`, the following is output:

```
lxd_container.v1: Creating...
╷
│ Error: Could not upload file /scripts/docker-setup.sh: Instance not found
│
│   with lxd_container.v1,
│   on main.tf line 47, in resource "lxd_container" "v1":
│   47: resource "lxd_container" "v1" {

```
It works correctly if I comment `type = "virtual-machine"` from the container resource

tarruda:
Creating a new instance that has `file {` blocks doesn't work if the type is "virtual-machine" because it relies on LXD Agent running.

Modify resourceLxdInstanceCreate to ensure the LXD Agent is running before trying to upload files. All other cases remain unmodified, and the user is responsible for ensuring the VM is running when making changes.

Close #315

### DIFF
--- a/docs/resources/instance.md
+++ b/docs/resources/instance.md
@@ -145,6 +145,21 @@ The `file` block supports:
 * `create_directories` - *Optional* - Whether to create the directories leading
 	to the target if they do not exist.
 
+Note that the way file upload is handled differs if the instance is a virtual machine:
+
+With containers, the file is copied directly to the container filesystem by
+LXD, and this can happen even if the container is not running.
+
+With virtual machines, the LXD agent must be running which requires the VM to
+be booted.
+
+For this reason, specifying `file` on instances of type "virtual-machine" will
+force the instance to be booted on creation, and the upload waits until the LXD
+agent is running.
+
+When updating an instance, the user must ensure the VM is running before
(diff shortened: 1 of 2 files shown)


tarruda:
Closing since "lxd_instance_file" can be used with "depends_on" to achieve the same effect as what this PR is doing.

adamcstephens:
Thanks for attempting to fix this, but I agree using the other resource is probably better.

tarruda:
@adamcstephens I'm thinking, since `lxd_instance_file` obviously depends on an instance, should we add an implicit "depends_on" (if such thing is possible)?

PR #323, opened 2023-08-03
README: explain how to use the provider built from sources

simondeziel:
After struggling to make use of the provider as built from sources, here's the instructions I came up with the help of @MusicDin. I'm happy to refine the doc if you have a better development workflow to suggest. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,26 @@ terraform {
 }
 ```
 
-### Building from source
+### Using the terraform-provider-lxd from source
 
 1. Follow these [instructions](https://golang.org/doc/install) to setup a Golang development environment.
-2. Use `go install` to compile and install the binary the binary:
+2. Checkout the repository `git clone ...`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,13 +28,30 @@ terraform {
...
+```shell
+$ cat ~/.terraformrc
+provider_installation {
+  dev_overrides {
+    "terraform-lxd/lxd" = "/home/<REPLACE_ME>/git/terraform-provider-lxd"
+  }
+
+  direct {
+    exclude = ["terraform-lxd/lxd"]
+  }
```
adamcstephens, line 53: What does this do? I haven't needed it in my testing.

simondeziel: I just tested without and indeed, that doesn't seem to be needed, thanks! I dropped it now.

adamcstephens (approved):

PR #324, opened 2023-08-04
Search for certificates in snap's config directory

MusicDin:
Currently, the provider uses `~/.config/lxc` directory for certificates, however, @simondeziel has pointed out that the provider should also check `~/snap/lxd/common/config` directory, which is used when the LXD is installed using Snap.

We propose, that the provider first checks whether the `/var/snap/lxd` directory exists. If it does, the provider should use `~/snap/lxd/common/config`, or `~/.config/lxc` otherwise.

Any thoughts on that?


adamcstephens:
The selection behavior seems reasonable to me. Is this the best way to check if lxd is installed through snap?

simondeziel:
> Is this the best way to check if lxd is installed through snap?

Checking the existence of `/var/snap/lxd` should be quick and reliable.

lxd/provider.go
```diff
@@ -160,7 +160,16 @@ func Provider() *schema.Provider {
...
+					// Check for the presence of the /var/snap/lxd directory. If the
+					// directory exists, return snap's config path, otherwise return
+					// the fallback path.
+
+					_, err := os.Stat("/var/snap/lxd")
+					if os.IsNotExist(err) {
+						return os.ExpandEnv("$HOME/.config/lxc"), nil
+					}
+
+					return os.ExpandEnv("$HOME/snap/lxd/common/config"), nil
```
simondeziel, line 172: The comment doesn't match the code. Checking for the dir first should lead to a smaller diff like:

```
    // Check for the presence of the /var/snap/lxd directory. If the
    // directory exists, return snap's config path, otherwise return
    // the fallback path.

    _, err := os.Stat("/var/snap/lxd")
    if os.IsExist(err) {
	return os.ExpandEnv("$HOME/snap/lxd/common/config"), nil
    }

    return os.ExpandEnv("$HOME/.config/lxc"), nil
```

lxd/provider.go
```diff
@@ -217,7 +226,7 @@ var descriptions map[string]string
 func init() {
 	descriptions = map[string]string{
 		"lxd_accept_remote_certificate":    "Accept the server certificate",
-		"lxd_config_dir":                   "The directory to look for existing LXD configuration. default = $HOME/.config/lxc",
+		"lxd_config_dir":                   "The directory to look for existing LXD configuration. default = $HOME/snap/lxd/common/config",
```
simondeziel, line 229: Dunno if that make sense but we could use a similar notation as for `$PATH` in `shell` so `default = $HOME/snap/lxd/common/config:$HOME/.config/lxc"`

lxd/provider.go
MusicDin: I agree, will fix that one.

MusicDin: Just to clarify the change, the `os.IsExist` is not sufficient, since this checks whether the error is of type *file already exists* or *directory not empty*. If there is no error, `os.IsExist` will return false, so we also have to check whether the `err == nil`.

simondeziel: Ah, right, thanks!

adamcstephens (approved):
Thanks!

PR #325, opened 2023-08-04
cached_image: add virtual-machine image alias support

### LINKED ISSUE #257: lxd_cached_image support of "virtual-machine" type image?
Currently LXD brings VM type image support with new attribute "Type" which has two possible values: `["virtual-machine", "container"]`. On CLI, we pull VM image with additional argument `--vm`.

Without support of this new attribute, every pull will just download **container** type image to local. Adding a `type` argument should provide the support of cache VM type image.

Ideally adding a `fingerprint` argument  will let user specify the specific version of image to download. Or in situation of not having `type` argument in place, people can specify a fingerprint of a VM image to make `lxd_cached_image` download it.

A theoretical resource definition can look like this:
```hcl
resource "lxd_cached_image" "ubuntu20vm" {
  name = "ubuntu20vm"
  source_remote = "ubuntu"
  source_image  = "focal/amd64"
  aliases = [ "ubuntu20vm" ]
  type = "virtual-machine"
  fingerprint = "9b989e450514"
}
```

Thanks

adamcstephens:
resolves #257

### DIFF
--- a/docs/resources/cached_image.md
+++ b/docs/resources/cached_image.md
@@ -27,6 +27,9 @@ resource "lxd_instance" "test1" {
 
 * `source_image` - *Required* - Fingerprint or alias of image to pull.
 
+* `type` - *Optional* - Type of image to cache. Must be one of `container` or `virtual-machine`.
+  Defaults to `container`.
+
 * `aliases` - *Optional* - A list of aliases to assign to the image after
 	pulling.
 
--- a/lxd/resource_lxd_cached_image.go
+++ b/lxd/resource_lxd_cached_image.go
@@ -54,6 +54,21 @@ func resourceLxdCachedImage() *schema.Resource {
 				Default:  "",
 			},
 
+			"type": {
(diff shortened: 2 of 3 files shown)


simondeziel (approved):
Just test it and it works well, thank you!

docs/resources/cached_image.md
```diff
@@ -27,6 +27,8 @@ resource "lxd_instance" "test1" {
 
 * `source_image` - *Required* - Fingerprint or alias of image to pull.
 
+* `type` - *Optional* - Type of image to cache. Must be one of `container` or `virtual-machine`
```
MusicDin, line 30: I would just emphasize here that it defaults to `container`.

adamcstephens: Thanks. I noticed this yesterday while reviewing again on my phone and is actually what stopped me from clicking the merge button. :)

MusicDin (approved):

PR #326, opened 2023-08-08
Support config.cloud-init keys

simondeziel:
This avoid triggering repeated changes on every run even when the `config.cloud-init` keys do not change.

### DIFF
--- a/lxd/resource_lxd_instance_test.go
+++ b/lxd/resource_lxd_instance_test.go
@@ -163,6 +163,20 @@ func TestAccInstance_updateConfig(t *testing.T) {
 					testAccInstanceConfig(&instance, "user.user-data", "#cloud-config"),
 				),
 			},
+			{
+				Config: testAccInstance_updateConfig3(instanceName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccInstanceRunning(t, "lxd_instance.instance1", &instance),
+					resource.TestCheckResourceAttr("lxd_instance.instance1", "name", instanceName),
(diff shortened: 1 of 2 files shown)


lxd/resource_lxd_instance_test.go
```diff
@@ -892,6 +906,21 @@ resource "lxd_instance" "instance1" {
...
 
+func testAccInstance_updateConfig3(name string) string {
+	return fmt.Sprintf(`
+resource "lxd_instance" "instance1" {
+  name = "%s"
+  image = "images:alpine/3.16"
+  profiles = ["default"]
+  config = {
+    "cloud-init.vendor-data" = "#cloud-config"
+	"user.dummy" = 5
```
MusicDin, line 917: This config key seems to be indented a bit too much.

lxd/resource_lxd_instance_test.go
```diff
@@ -892,6 +906,21 @@ resource "lxd_instance" "instance1" {
...
+func testAccInstance_updateConfig3(name string) string {
+	return fmt.Sprintf(`
+resource "lxd_instance" "instance1" {
+  name = "%s"
+  image = "images:alpine/3.16"
+  profiles = ["default"]
+  config = {
+    "cloud-init.vendor-data" = "#cloud-config"
+    "user.dummy" = 5
+    "user.user-data" = "#cloud-config"
```
adamcstephens, line 918: Should we test with `cloud-init.user-data` instead since `user.user-data` is documented as legacy?

simondeziel:
Let's wait on the API feature detection bits to land to be able to do the right thing in tests.

PR #327, opened 2023-08-09
Fix not found error checks

MusicDin:
Currently, when the provider retrieves LXD storage pool (or tries to delete it), it checks whether an error equals to `No such object`. In such case it detects that the storage pool does not exist and an error is not reported.

However, LXD returns different error in such cases (https://github.com/canonical/lxd/blob/7f46f71133411a66be20fce1a2b92de3923db705/lxd/db/storage_pools.go#L666) - `Storage pool not found`.

### DIFF
--- a/lxd/errors.go
+++ b/lxd/errors.go
@@ -2,8 +2,20 @@ package lxd
 
 import (
 	"errors"
+	"net/http"
+
+	"github.com/canonical/lxd/shared/api"
 )
 
 var errNetworksNotImplemented = errors.New("This LXD server does not support " +
 	"the creation of networks. You must be running LXD 2.3 or later for this " +
 	"feature.")
+
+func isNotFoundError(err error) bool {
+	// For LXD versions below "4.0".
+	if err.Error() == "No such object" || err.Error() == "not found" {
+		return true
+	}
+
+	return api.StatusErrorCheck(err, http.StatusNotFound)
+}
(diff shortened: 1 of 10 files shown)


MusicDin:
Since the error message has been changed only one year ago (https://github.com/canonical/lxd/commit/ce7b3fb2692765c26520a46ea0b91872ca0a7ede), maybe we should keep both options for now (for those who are still running older versions of LXD):
```go
if err != nil && (err.Error() == "No such object" || err.Error() == "Storage pool not found") {
        // ...
}
```

@simondeziel @adamcstephens Any thoughts on that?

simondeziel:
Thanks for thinking about backward compat! The commit you found was backported to 5.0 LTS (https://github.com/canonical/lxd/commit/d442d204d4677a4ae0efc82e8e0f4964c67af249) but not to 4.0 LTS, so yeah, I think it's worth checking for both error strings.

Excuse my ignorance but isn't there a HTTP return code accompanying the error? I'd expect the old and new errors to both come with a `404`. If that's available, that'd potentially be less brittle than trying to look at the error string itself.

MusicDin:
You are right, we could check if the error is of type `api.StatusError` and extract the status code from it.

MusicDin:
I have replaced the string comparison error checks with the status code error checks. 

This approach works both for LXD version `4` and `5`. However, LXD versions `3` and below do not implement `StatusError`, which is why I have created an utility function in `errors.go` (maybe not an appropriate place for that) that checks if the error message equals `No such object` or `not found` before trying to convert it into StatusError.

In addition, I have replaced most of the `not found` errors which follow the same approach.

adamcstephens (approved):
Thanks!

PR #335, opened 2023-08-11
Add acceptance test prechecks 

MusicDin:
This PR adds 4 acceptance test prechecks that skip the test execution if certain condition is not satisfied.
- `testAccPreCheckLxdVersion` skips the test if LXD does not satisfy the version constraint.
- `testAccPreCheckAPIExtensions` skips the test if LXD does not support all the required extensions.
- `testAccPreCheckClustering` skips the test if LXD does not run in a clustered mode.
- `testAccPreCheckVirtualization` skips the test if LXD does not support virtualization.

This is required by @simondeziel to be able to properly configure tests for older LXD versions.


adamcstephens (approved):
utils.go is not my favorite, but I have no better ideas.

Thanks! 🍕

PR #337, opened 2023-08-16
Use unix as default remote scheme instead of https

MusicDin:
Currently, [provider's docs](https://registry.terraform.io/providers/terraform-lxd/lxd/latest/docs#scheme) state that if remote scheme is not set, `unix` is used by default, which is not true. 

This PR changes the default remote scheme from `https` to `unix`. However, we can also fix the docs instead.

@adamcstephens What are your thoughts?


adamcstephens (approved):
Thanks 🍕

PR #338, opened 2023-08-24
Add LXD version constraint

### LINKED ISSUE #333: Drop support for LXD versions lower then `3.0`
I've been discussing with @simondeziel and we thought that it could be beneficial if we define a set of supported LXD versions. 

Considering LXD `3.x` has recently reached end-of-life, it seems practical to no longer support version `2.x`. By focusing on the three most recent LXD versions, we can remove the long time deprecated code, while still providing users a reasonable window for upgrades. 

Also, if someone's using the oldest version supported by the provider (for example `3.x`), we could issue a Terraform warning. This would inform the users that their current version is nearing deprecation, ensuring they aren't caught off-guard by future updates.

Any thoughts or suggestions?

MusicDin:
This PR sets the minimum LXD version (`3.0.0`) supported by the provider.
Currently, only the default remote is verified.

Fixes #333

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -7,13 +7,15 @@ Use Terraform to manage LXD resources.
 This provider connects to the LXD daemon over local Unix socket or HTTPS.
 
 It makes use of the [LXD client library](https://github.com/canonical/lxd), which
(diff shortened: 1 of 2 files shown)


adamcstephens (approved):
Thanks! 🍕

PR #339, opened 2023-08-24
Use test precheck to detect if LXD is in clustered mode

MusicDin:
Use existing test prechecks for detecting whether LXD is in clustered mode.


adamcstephens (approved):
Thanks!

PR #346, opened 2023-09-07
Add support for network zones

mandrav:
This patch add support for lxd_network_zone and lxd_network_zone_record resources.


MusicDin:
I've quickly went through the code, and we should check whether the target LXD server supports the appropriate extension.

Something like this:
```go
nzExt := "projects_networks_zones"
if !server.HasExtension(nzExt) {
	return fmt.Errorf("Network zones cannot be used because LXD server does not support %q extension.", nzExt)
}
```

However, I'm not sure `projects_networks_zones` is the appropriate one.
@simondeziel could you help here?

docs/resources/resource_lxd_network_zone.go
adamcstephens, line 1: This go file shouldn't be here.

adamcstephens (changes requested):
I haven't fully reviewed this yet, but you're missing the documentation for the zone and tests for the record. Also note the CI warnings/errors.

simondeziel:
> I've quickly went through the code, and we should check whether the target LXD server supports the appropriate extension.
> 
> Something like this:
> 
> ```go
> nzExt := "projects_networks_zones"
> if !server.HasExtension(nzExt) {
> 	return fmt.Errorf("Network zones cannot be used because LXD server does not support %q extension.", nzExt)
> }
> ```
> 
> However, I'm not sure `projects_networks_zones` is the appropriate one. @simondeziel could you help here?

tl;dr: you've got the right extension IMHO.

The DNS support depends on `network_dns_records` but when only that extension is present, the network zone is created in the same project as the network it is associated with. The `projects_networks_zones` extension decouples the 2 which this PR seems to be using so I think only checking for `projects_networks_zones` is good.

lxd/resource_lxd_network_zone.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+}
+
+func resourceLxdNetworkZoneCreate(d *schema.ResourceData, meta interface{}) error {
+	p := meta.(*lxdProvider)
+	server, err := p.GetInstanceServer(p.selectRemote(d))
+	if err != nil {
+		return err
+	}
+
```
MusicDin, line 60: Can you please just add a check if server supports `projects_networks_zones` extension:

```go
netZoneExt := "projects_networks_zones"
if !server.HasExtension(netZoneExt) {
	return fmt.Errorf("Network zones cannot be used because LXD server does not support %q extension", netZoneExt)
}
```

lxd/resource_lxd_network_zone.go
```diff
@@ -0,0 +1,241 @@
...
+	return resourceLxdNetworkZoneRead(d, meta)
+}
+
+func resourceLxdNetworkZoneRead(d *schema.ResourceData, meta interface{}) error {
+	p := meta.(*lxdProvider)
+	server, err := p.GetInstanceServer(p.selectRemote(d))
+	if err != nil {
+		return err
+	}
+
```
MusicDin, line 94: Same comment as above. I think adding extension check to Read and Create should be sufficient.

MusicDin (changes requested):

lxd/resource_lxd_network_zone.go
mandrav: Done.

docs/resources/network_zone_record.md
```diff
@@ -0,0 +1,72 @@
...
+
+  config = {
+    "dns.nameservers"  = "ns.custom.example.org"
+    "peers.ns.address" = "127.0.0.1"
+  }
+}
+
+resource "lxd_network_zone_record" "record" {
+  name = "ns"
+  zone = lxd.network_zone.zone.id
```
MusicDin, line 25: Just a small typo here: `lxd.network_zone` -> `lxd_network_zone`

docs/resources/network_zone.md
```diff
@@ -0,0 +1,65 @@
...
+
+  config = {
+    "dns.nameservers"  = "ns.custom.example.org"
+    "peers.ns.address" = "127.0.0.1"
+  }
+}
+
+resource "lxd_network_zone_record" "record" {
+  name = "ns"
+  zone = lxd.network_zone.zone.id
```
MusicDin, line 25: Just a small typo here: `lxd.network_zone` -> `lxd_network_zone`

MusicDin (changes requested):

MusicDin (approved):

adamcstephens (approved):
LGTM. Thanks for your contribution! :pizza:

PR #348, opened 2023-09-07
Drop lxd_volume_container_attach resource

simondeziel:
This resource was deprecated in 2017 so it feels like a good time to drop it.


adamcstephens:
I think we'd want to cut a major release to remove a deprecated resource. (Though it has been quite a while) It looks like LTS 3.0 is no longer supported upstream, so maybe we can batch up a few breaking changes and release 2.0? It's probably too soon to drop the `container` resources, but if y'all think it should be included I'd be happy to discuss.

MusicDin:
Is this deprecated in LXD `3.0.0`? Otherwise we could wait until we drop support for this LXD version

simondeziel:
LXD 3.0 is the last release that does not support VMs.

simondeziel:
In terms of breaking changes leading to `terraform-provider-lxd` 2.0, I'd like to drop:

* `lxd_volume_container_attach` (this PR)
* `lxd_container` and `lxd_container_file`
* the provider `address`, `scheme`, `remote`, etc fields replaced by their `lxd_remote.*` equiv

simondeziel:
To provide more context, LXD 3.0 was released with Ubuntu 18.04 so back in 2018. It went out of regular support in April 2023. Even if you still have some 18.04 kicking around, you can switch to the snap and migrate to LXD 4.0 easily. As such, I'm all for making 4.0 the minimum supported version.

MusicDin:
If I understand correctly, this `attach` resource can no longer be used as it was no longer supported in `3.0.0`? If so, we can safely merge this PR, as user already cannot use this provider with LXD lower then 3.0.0.

For next major release I would also like to finish configuration management, as currently it is broken wherever LXD uses configuration map (network and storage being most critical).

simondeziel:
> If I understand correctly, this `attach` resource can no longer be used as it was no longer supported in `3.0.0`? If so, we can safely merge this PR, as user already cannot use this provider with LXD lower then 3.0.0.

No, the 2 are unrelated. The `lxd_volume_container_attach` resource is not needed as you can accomplish the same with an instance device referencing a `lxd_volume`.

> For next major release I would also like to finish configuration management, as currently it is broken wherever LXD uses configuration map (network and storage being most critical).

Indeed, that'd nice, I only provided a "legacy code to drop" wishlist ;)

adamcstephens:
Are y'all interested in breaking out your wishlists into issues and linking to https://github.com/orgs/terraform-lxd/projects/1 ? We can also just use a tracking issue if preferred.

I'll be out of town the next couple days, but I'd love to work out what we want to clean up and track towards a new release. Please share any thoughts you have. :)

MusicDin:
I'm currently migrating to terraform-framework and am wondering whether we should separate resource into different directories (for readability)?

adamcstephens:
> I'm currently migrating to terraform-framework and am wondering whether we should separate resource into different directories (for readability)?

That works for me. I have no attachment to the current flat structure. Is there a recommendation either way with terraform-framework?

adamcstephens (approved):

simondeziel:
@adamcstephens I'll leave it up to you to review/merge the few PRs that are on the `provider 2.0` release roadmap.

PR #349, opened 2023-09-07
Add `remote` attribute to `lxd_project` and deprecate `target`

simondeziel:


MusicDin (changes requested):
Could you please update the documentation for `lxd_project`?

lxd/resource_lxd_project.go
```diff
@@ -32,10 +32,18 @@ func resourceLxdProject() *schema.Resource {
...
 				ForceNew: true,
+				Optional: true,
+				Default:  "",
+			},
+
+			"target": {
+				Type:       schema.TypeString,
+				Optional:   true,
+				ForceNew:   false,
+				Deprecated: "This attribute is ignored",
```
MusicDin, line 46: Is this still necessary if target is no longer used?

MusicDin (approved):
LGTM, let's wait for @adamcstephens to see if we have missed something.

PR #355, opened 2023-10-10
Determine LXD socket before retrieving a client

MusicDin:
Instead of using LXD socket in snap's directory when an error contains `/var/lib/lxd/unix.socket` in the error, determine the LXD socket before retrieving the client.

### DIFF
--- a/lxd/provider.go
+++ b/lxd/provider.go
@@ -6,7 +6,6 @@ import (
 	"log"
 	"os"
 	"path"
-	"strings"
 	"sync"
 	"time"
 
@@ -15,6 +14,7 @@ import (
 	"github.com/canonical/lxd/shared"
 	lxd_api "github.com/canonical/lxd/shared/api"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"golang.org/x/sys/unix"
 )
 
 // A global mutex.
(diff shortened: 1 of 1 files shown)


MusicDin:
@simondeziel @adamcstephens What do you think if we set `TerraformLxdConfig` for `local` during provider's initialization? This way there would be no need to set `LXD_DIR`/`LXD_SOCKET` environment variable before retrieving a client. However, I am not sure if this would have any negative implications on existing codebase..

lxd/provider.go
```diff
@@ -766,3 +757,39 @@ func determineDaemonAddr(lxdRemote terraformLXDConfig) (string, error) {
...
+func determineLxdDir() (string, error) {
+	lxdSocket, ok := os.LookupEnv("LXD_SOCKET")
+	if ok {
+		return path.Dir(lxdSocket), nil
+	}
+
+	lxdDir, ok := os.LookupEnv("LXD_DIR")
+	if ok {
+		return lxdDir, nil
+	}
```
adamcstephens, line 773: Should we also check this directory for a writable socket, or just assume the user knows what they're doing?

MusicDin: Good point, we could check both LXD_SOCKET and LXD_DIR (if any of those is set) to fail faster. I'll fix that.

simondeziel:
> @simondeziel @adamcstephens What do you think if we set `TerraformLxdConfig` for `local` during provider's initialization? This way there would be no need to set `LXD_DIR`/`LXD_SOCKET` environment variable before retrieving a client. However, I am not sure if this would have any negative implications on existing codebase..

I don't see any harm in exporting those 2 env variables.

simondeziel (approved):

adamcstephens:
This isn’t Windows compatible so will need to be refactored. It’s blocking the release. https://github.com/terraform-lxd/terraform-provider-lxd/actions/runs/6698055575/job/18200325673

PR #362, opened 2023-10-27
Fix ignored error in instance/container exists function

MusicDin:
Error was ignored in instance/container exists function.

While this fixes the issue, we should probably get rid of named returned values wherever possible to prevent such issues in the future. I think something like this is even more readable:
```go
_, _, err := server.GetInstanceState(name)
if err != nil {
         if isNotFoundError(err) {
                 return false, nil
         }
         
         return false, err
}

return true, nil
```

Any thoughts?

### DIFF
--- a/lxd/resource_lxd_container.go
+++ b/lxd/resource_lxd_container.go
@@ -781,7 +781,7 @@ func resourceLxdContainerExists(d *schema.ResourceData, meta interface{}) (exist
 		exists = true
 	}
 
-	return exists, nil
+	return
 }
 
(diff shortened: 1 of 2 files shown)


adamcstephens:
What is the problem you're trying to solve? I actually added these explicit returns because previously externally deleted instances would cause a hard failure. https://github.com/terraform-lxd/terraform-provider-lxd/pull/298

I don't like the named returns either, and prefer to be explicit with the returns.

MusicDin:
Ahh, I see. Thought the error is ignored because we are returning an error for all other resources.

Let's avoid them in the future and gradually get rid of the existing ones.

PR #364, opened 2023-10-30
testing: cleanup and catch more

adamcstephens:
reduce duplicate test runs, run more in parallel, and test building for all release platforms

build-platforms test will fail, given that Windows is already failing.


adamcstephens:
Any concerns? I'll merge this broke and we can open another PR to fix the failing build.

MusicDin (approved):
LGTM

I'll open temporary fix for Win failure.

PR #372, opened 2023-11-13
test: Fix instance reference within test

MusicDin:


simondeziel (approved):

adamcstephens:
Aha, I spent some time last week and couldn't track this down. Thanks for finding it. I'm not sure how this ever passed.

MusicDin:
Yes, this is exactly what I don't understand. While this fixed the test that is running locally, GH actions log indicates VM tests were skipped:
- https://github.com/terraform-lxd/terraform-provider-lxd/actions/runs/6850795786/job/18625683766#step:8:122
- https://github.com/terraform-lxd/terraform-provider-lxd/actions/runs/6850795786/job/18625683766#step:8:69

adamcstephens:
That's not good. I suspect some actions runners don't support virtualization while some do. And others appear to be seeing the same https://github.com/actions/runner-images/discussions/7191#discussioncomment-7548174

It's a shame we can rely on these tests. I'd suggest we fail the build if virtualization is missing, but then there's no reliable way to make sure we do get a runner that supports it.

MusicDin:
Hmmm, let's make sure all VM tests are passing locally. I'm experiencing some issues with old container resource where `GET /v1/containers` cannot find the instance, but I can see it running in LXD.

I'm assuming the tests will pass then even when virtualization is enabled on GH runner.

MusicDin:
Seems that virtualization works when enabled on GH runner: https://github.com/terraform-lxd/terraform-provider-lxd/actions/runs/6851234379/job/18627079116#step:8:122

PR #373, opened 2023-11-13
test: Drop VM test for old lxd_container resource

MusicDin:
~This is a bit of a workaround, but seems that client's `GetContainer()` is incapable of properly returning VM instances.~

~We can either drop the VM test for old container resource, or rely on `GetInstance()` to verify the container is running instead.~

```
$ lxc launch ubuntu:22.04 v1 --vm
$ lxc launch ubuntu:22.04 c1

$ lxc query -X GET /1.0/containers
[
        "/1.0/containers/c1"
]

$ lxc query -X GET /1.0/instances
[
        "/1.0/instances/c1",
        "/1.0/instances/v1"
]
```

EDIT:

This PR just removes VM test for `lxd_container` resource. Resource `lxd_container` will be dropped soon and VMs are also tested for `lxd_instance` resource.

### DIFF
--- a/lxd/resource_lxd_container_test.go
+++ b/lxd/resource_lxd_container_test.go
@@ -68,25 +68,6 @@ func TestAccContainer_typeContainer(t *testing.T) {
 	})
 }
 
-func TestAccContainer_typeVirtualMachine(t *testing.T) {
-	var container api.Container
-	containerName := petname.Generate(2, "-")
-
-	resource.Test(t, resource.TestCase{
-		PreCheck:  func() { testAccPreCheckVirtualization(t) },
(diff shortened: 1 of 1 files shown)


MusicDin:
Seems that `/1.0/containers` does not return VMs. @simondeziel pointed me to this doc section: https://documentation.ubuntu.com/lxd/en/latest/rest-api/#instances-containers-and-virtual-machines

simondeziel:
> We can either drop the VM test for old container resource

With the confirmation that `/1.0/containers` never had/never will show VMs, it should be safe to drop it.

simondeziel (approved):

PR #375, opened 2023-11-13
Test for required API extensions in network LB and zones tests

simondeziel:
This is in preparation of supporting LXD `5.0/stable`.


MusicDin (approved):
LGTM

PR #379, opened 2023-11-14
Make LXD 4.0 the minimum officially supported version

simondeziel:
LXD 4.0 is the oldest currently supported branch by LXD upstream.

### DIFF
--- a/lxd/errors.go
+++ b/lxd/errors.go
@@ -1,17 +1,11 @@
 package lxd
 
 import (
-	"errors"
 	"net/http"
 
 	"github.com/canonical/lxd/shared/api"
 )
 
 func isNotFoundError(err error) bool {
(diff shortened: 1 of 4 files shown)


MusicDin (approved):
LGTM

I think we can also cleanup `isNotFoundError` in errors.go. I think from LXD 4.0 there is no longer a need to check for "not found" in an error string

simondeziel:
> I think we can also cleanup `isNotFoundError` in errors.go. I think from LXD 4.0 there is no longer a need to check for "not found" in an error string

I've removed the error string comparison bit of that function. I've also confirmed that LXD 4.0 was indeed returning a 404 error.

adamcstephens (approved):
Thanks! :pizza:

PR #386, opened 2023-11-30
Add exec block to the instance resource

### LINKED ISSUE #54: how to perform "remote-exec", but inside the LXD
Terraform version: Terraform v0.9.6
terraform-provider-lxd: 0.9.5 beta1

As part of the LXD creation, I push some config files and a service startup script (upstart)
These cannot be baked into the LXD image.

It would be quite useful if I could target command execution directly inside the LXD container.

For now, I am making do with a "remote-exec" on the LXD server and do:
```lxd exec mylxd1 -- telinit 2```  (to make the upstart script start)

Regards,
Shantanu

### PR DESCRIPTION
**Pull request by MusicDin**
This PR adds `exec` block to the instance resource which allows running commands within the instance.

- [x] Requires #385 to be rebased.
- [x] Requires #387 to be able to properly manage `exec` state.

Fixes #54

### DIFF
--- a/docs/resources/instance.md
+++ b/docs/resources/instance.md
@@ -156,6 +156,9 @@ The `exec` block supports:
 * `record_output` - *Optional* - When set to true, `stdout` and `stderr` attributes will be
   populated (exported). Defaults to `false`.
 
+* `fail_on_error` - *Optional* - Boolean indicating whether resource provisioning should stop upon
+  encountering an error during command execution. Defaults to `false`.
+
 * `uid` - *Optional* - The user ID for running command. Defaults to `0` (root).
 
 * `gid` - *Optional* - The group ID for running command. Defaults to `0` (root).
@@ -242,8 +245,10 @@ resource "lxd_instance" "inst" {
 
 ### Capturing Command Output
 
-To capture and access a command's output, set `record_output` to true. The command's standard output
-and standard error will then be accessible through the exported attributes `stdout` and `stderr`, respectively.
+Exit status of the command will be always available after command execution via `exit_code` attribute.
+However, to capture and access a command's output, set `record_output` to true. The command's standard
+output and standard error will then be accessible through the exported attributes `stdout` and `stderr`,
+respectively.
 
 ```hcl
 resource "lxd_instance" "inst" {
@@ -258,8 +263,26 @@ resource "lxd_instance" "inst" {
 
 output "exec-output" {
   value = {
-    "out" = lxd_instance.inst.exec[0].stdout # "Linux\n"
-    "err" = lxd_instance.inst.exec[0].stderr # ""
+    "code" = lxd_instance.inst.exec[0].exit_code # 0
+    "out"  = lxd_instance.inst.exec[0].stdout    # "Linux\n"
+    "err"  = lxd_instance.inst.exec[0].stderr    # ""
+  }
+}
+```
+
+### Fail on Command Error
+
+By default, command failure is ignored. If you want to stop Terraform from provisioning the resources
+if command exits with a non 0 status, set `fail_on_error` attribute to true.
+
+```hcl
+resource "lxd_instance" "inst" {
+  name      = "c1"
+  image     = "images:alpine/3.18/amd64"
+
+  exec {
+    command       = ["invalid-cmd"]
+    fail_on_error = true
   }
 }
 ```
--- a/docs/resources/instance.md
+++ b/docs/resources/instance.md
@@ -201,12 +201,13 @@ resource "lxd_instance" "instance1" {
 
 The `exec` block in an LXD instance configuration is used to execute commands. You can specify
 multiple exec blocks, with each block requiring a command defined as a list of strings.
+It's important to note that commands will be executed in the same order as the exec blocks appear
+in the configuration.
 
-### Simple Commands
+### Commands and Environment Access
 
-For simple and short commands, you can specify the entire command as a single string in the list.
-In this case, if the list contains only one string, the provider automatically splits it into
-multiple arguments based on spaces.
+The command should be specified as a list of strings, where the first string is the path to
+an executable, followed by its arguments.
 
 ```hcl
 resource "lxd_instance" "inst" {
@@ -214,14 +215,14 @@ resource "lxd_instance" "inst" {
   image = "images:alpine/3.18/amd64"
 
   exec {
-    command = ["ls -lah"]
+    command = ["ls", "/"]
   }
 }
 ```
 
-### Complex Commands and Environment Access
-
-For more complex commands or when access to the environment is required, use the `<shell> -c` syntax.
+By default, the command executes directly without a shell, meaning shell patterns like variables
+and file redirects are not interpreted. To use environment variables or shell meta-characters
+(e.g., pipes and file redirects), you must specify a shell executable.
 
 ```hcl
 resource "lxd_instance" "inst" {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by MusicDin** [comment]
@simondeziel Your comment from #387 PR:

> One last thing regarding the "working devlxd communication" aspect. How does that degrades if someone has security.devlxd=false? For containers, this is perfectly usable but I could see someone using it on VMs too where to connection mean is either the lxc console or ssh.

Just to not forget about it.

**Comment by simondeziel** [comment]
> @simondeziel Your comment from #387 PR:
> 
> > One last thing regarding the "working devlxd communication" aspect. How does that degrades if someone has security.devlxd=false? For containers, this is perfectly usable but I could see someone using it on VMs too where to connection mean is either the lxc console or ssh.
> 
> Just to not forget about it.

To clarify, I was wrong and `security.devlxd=false` does not affect the ability to `exec` into VMs.

**Review by simondeziel — COMMENTED** [review]
I like the proposal but have a few comments (as always ;)

I think Puppet really provides an easy to work with `exec` resource that we should take inspiration from.

File: internal/instance/resource_instance_test.go [hunk 1]
```diff
@@ -1132,6 +1261,84 @@ resource "lxd_instance" "instance1" {
...
+}
+
+func testAccInstance_execWorkingDir(instanceName string) string {
+	return fmt.Sprintf(`
+resource "lxd_instance" "instance1" {
+  name  = "%s"
+  image = "%s"
+
+  exec {
+    command       = ["sh", "-c", "cat os-release | grep '^ID' | tr -d '\n'"]
```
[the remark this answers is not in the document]
MusicDin: This is supposed to test whether special characters such as backslash (`\`) and pipe (`|`) are correctly passed to the instance. But yeah, could be for sure something something more readable :sweat_smile:

internal/instance/resource_instance.go
```diff
@@ -286,6 +288,78 @@ func (r InstanceResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+							Optional: true,
+							Computed: true,
+							Default:  stringdefault.StaticString(""),
+							Validators: []validator.String{
+								stringvalidator.LengthAtLeast(1),
+							},
+						},
+
+						"record_output": schema.BoolAttribute{
+							Description: "Whether to record command's standard output and error (Setting this to true populates stdout and stderr attributes)",
```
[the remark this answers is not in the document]
MusicDin: Good idea. Will add, thanks.

docs/resources/instance.md
```diff
@@ -141,6 +141,25 @@ The `file` block supports:
 * `create_directories` - *Optional* - Whether to create the directories leading
 	to the target if they do not exist.
 
+The `exec` block supports:
+
+* `command` - **Required** - The command to be executed (list of strings).
+
+* `triggers` - *Optional* - A list of arbitrary strings that, when changed, will force the command
+  to be rerun.
```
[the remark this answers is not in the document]
MusicDin: To recap our yesterday's conversation. The tricky part about special conditions is that we need to evaluate the plan of an instance before the update without doing any modification on the infrastructure itself. This means that any checks that need to be executed within an instance, would require instance to be running, which we *can not* ensure during the planning phase. For VMs, even simple file require the instance to be running.

Technically, we could ensure the instance is running during the planning phase by simply starting it if it is stopped, but this goes against Terraform concepts because we are doing an *update* without prior approval from the user. In addition, planning phase could take up to a minute instead of a few seconds.

Another approach would be to treat those special conditions as *requires update*, which means that such instances would always show *plan changes*, even if the conditions are already satisfied. Conditions would then be evaluated only during an actual update.

However, running a command on each apply can be already achieved using triggers:
```hcl
exec {
  command  = ["..."]
  triggers = [timestamp()] # This will ensure the change is always detected.
}
```

With such approach we can add new conditions that will determine whether command needs to be run, but they should not trigger any plan changes, because we cannot verify whether the change is already satisfied.

For advance use cases, the easiest approach would be to upload Puppet/Ansible configuratio to an instance (using `file` block) and then apply it using `exec` block that has a `timestamp()` trigger.

internal/common/lxd_exec.go
```diff
@@ -39,9 +40,8 @@ func (e *ExecModel) Execute(ctx context.Context, server lxd.InstanceServer, inst
 		return diags
 	}
 
-	// If command is one liner, split it on spaces.
-	if len(command) == 1 {
-		command = strings.SplitN(command[0], " ", 3)
+	if e.Interpreter.ValueString() == "shell" {
+		command = append([]string{"/bin/sh", "-c"}, command...)
```
adamcstephens, line 44: I don't like hardcoding the interpreter to `/bin/sh`. It assumes that this path will exist, which is not necessarily true, and can be a different package depending on the image being worked with. Could we instead allow a user to provide their own interpreter rather than selecting a generic `shell` value?

MusicDin: Yeah, by default we do not use any interpreter, so the user can provide the command as:
```hcl
command = ["/bin/sh", "-c", "..."]
```

I though about having an interpreter as a list of strings (terraform's [local-exec](https://developer.hashicorp.com/terraform/language/resources/provisioners/local-exec#interpreter) uses that approach):
```hcl
command     = ["..."]
interpreter = ["/bin/sh", "-c"]
```

But it seems like just splitting a list into two lists, or? 

Could we maybe have `interpreter` attribute as you have proposed, but have `command` as a string? - However, how would then a simple command like `apt update` look like if we want to avoid using shell?
```hcl
command     = "apt update"
interpreter = ["sh", "-c"]

command     = "update"
interpreter = ["apt"]
```

Maybe the interpreter *should* be the required field?

simondeziel: If you take that route, might as well use this form:
```
command   = "apt" # note the bare string, not an array/list
arguments = ["dist-upgrade", "-V", "-y"]
```

adamcstephens: Some prior art would be the built-in local-exec provisioner. It sets a default of `/bin/sh -c ` (in a slice) if there is not a specified parameter and takes a simple string as the command. `working_dir` and `environment` may not be bad to offer either.

https://developer.hashicorp.com/terraform/language/resources/provisioners/local-exec
https://github.com/hashicorp/terraform/blob/main/internal/builtin/provisioners/local-exec/resource_provisioner.go

I'd be fine with defaulting to `/bin/sh` if there is an alternative

MusicDin: How about we have three attributes? - `posix`, `command`, `interpreter`

posix would be mutually exclusive with command and interpreter.
*I don't like posix as attribute name, but just for demo*

So user can provide either:
`posix = ["ls", "/"]` // So that we allow executing commands without shell

or:
```hcl
command     = "ls /"
interpreter = ["bash", "-c"] # Defaults to /bin/sh -c
```

With this approach, we could maybe also have list of `commands` instead of single command per exec block?

MusicDin: @terraform-lxd/maintainers Any comments/suggestions on https://github.com/terraform-lxd/terraform-provider-lxd/pull/386#discussion_r1421009176?

adamcstephens: I think my preference would be just using `command` and `interpreter`, with `interpreter` defaulting to `bash -c` and `command` being a string. This should support simple cases of just wanting to run a command and not worrying about the interpreter, but also allow for more complex cases where specifying the interpreter is desirable. I'm not sure if I see the value in also having a `posix` attribute, unless I'm missing something.

MusicDin: Thanks everyone. 

Based on the comments, I will leave `command` as is (list of strings), and change the `interpreter` attribute from string (enum) to list of strings. If interpreter is set, it will be simply prepended to the command.

simondeziel (approved):

PR #394, opened 2023-12-11
Remove refresh_interval attribute from provider

MusicDin:
Currently documentation for provider's `refresh_interval` states:
> How often to poll during state change.

However, this is a bit misleading. Refresh interval is used only within instance resource as `Delay` in `retry.StateChangeConf`.

[retry.StateChangeConf](https://github.com/hashicorp/terraform-plugin-sdk/blob/b48765e952eee6ff8d420ca7793325bffaea34de/helper/retry/state.go#L27C1-L39C2) has the following fields:
- `Timeout` - Maximum time to wait for the desired state.
- `MinTimeout` - Minimum time between checks, which keeps increasing by *times 2* with each unsuccessful wait up to 10 seconds. 
  - For example, if `MinTimeout = 2`, the polling would happen on 2, 4, 8, 10, 10, 10, ... seconds.
- `Delay` - Initial delay before the first check/refresh.
- `PollInterval` - Fixed polling interval (*currently not used*).

An instance with `running=true` and `wait_for_network=true` with delay of 10 seconds takes at least 20 seconds to be recognized as started.

This PR removes refresh_interval attribute and sets initial delay and minimum timeout to 2 seconds. Alternatively, we could use refresh_interval for *fixed* polling interval, if anyone sees use in that?

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -75,10 +75,6 @@ The following arguments are supported:
 	also be set with the `LXD_ACCEPT_SERVER_CERTIFICATE` environment variable.
   Defaults to `false`
 
-* `refresh_interval` - *Optional* - How often to poll during state change.
-	Defaults to "10s", or 10 seconds. Valid values are a Go-style parsable time
-	duration (`10s`, `1m`, `5h`).
-
 The `remote` block supports:
 
 * `address` - *Optional* - The address of the LXD remote.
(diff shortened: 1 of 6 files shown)


adamcstephens (approved):
This change makes sense to me. I don't see a need to keep this attribute. If anything, I would expect the Timeout to be a more useful knob to expose, and probably on the instance itself instead of globally. I vote for removing now and adding back if requested/needed.

MusicDin:
I've just rebased. It can be merged now.

adamcstephens (approved):

PR #397, opened 2023-12-13
Update go dependencies

MusicDin:
Update go dependencies. 

LXD is updated to release 5.19 (2023.10.19 09:52:45 [bad7f0beadba](https://github.com/canonical/lxd/commit/bad7f0beadba8c26eaa9329a5f67abb835d80004)) to avoid any potential issues with the recent license change.


adamcstephens:
Should we also add a replace to prevent someone from running `go get -u` accidentally upgrading this module?

I’m also a bit curious on how this will impact our ability to support lxd moving forward.

MusicDin:
> Should we also add a replace to prevent someone from running go get -u accidentally upgrading this module?

Nice suggestion, didn't even know that exists. I've tried that locally, but am not sure how `replace` is supposed to work?

```
replace github.com/canonical/lxd => github.com/canonical/lxd v0.0.0-20231019095245-bad7f0beadba

require (
	github.com/canonical/lxd v0.0.0-20231019095245-bad7f0beadba
	...
)
```

So if I run `go get -u`, it still replaces the required dependency with `github.com/canonical/lxd v0.0.0-20231212150554-000b1ab2a9e1`. Am I missing something? 

> I’m also a bit curious on how this will impact our ability to support lxd moving forward.

I've already inquired about this and will report back once its clarified. However, I would stick with 5.19 until then, just to be on the safe side.

simondeziel:
Incus' `lxd-to-incus` used a slightly different commit https://github.com/canonical/lxd/pull/12417/commits/ff5926c3b519757ba1aec04a253d3c3fb980f84b vs https://github.com/canonical/lxd/commit/bad7f0beadba8c26eaa9329a5f67abb835d80004 which is the "merge commit".

It's probably not relevant but it might save someone else the investigation if we used the same commit.

MusicDin:
> > I’m also a bit curious on how this will impact our ability to support lxd moving forward.

> I've already inquired about this and will report back once its clarified. However, I would stick with 5.19 until then, just to be on the safe side.

Third-party integrations should be allowed to use non-modified shared LXD packages without a need for re-licensing. So, nothing should change on our side in the future. 

However, since there are no changes in 5.20 that affect the provider, I propose we stick with 5.19 anyway, at least until this is officially clarified by Canonical.

MusicDin:
@simondeziel I've fixed the target commit as you have suggeseted.

simondeziel (approved):

PR #406, opened 2023-12-21
Switch away from `images:` and use `ubuntu-daily:` remote instead

simondeziel:
https://discuss.linuxcontainers.org/t/important-notice-for-lxd-users-image-server/18479

Also switches to using `ubuntu-minimal-daily:` for tests as those images are much smaller then regular ones from `ubuntu-daily:`.

Lastly, it replaces `candidate` channels by `edge` ones. The rational is to get an early warning from `edge` in case of problem and to have the assurance that we don't break users running `stable`.

Note that since those `-minimal` remotes are fairly recent, they are not yet available in `5.0/stable` so it's being temporarily disabled.


internal/acctest/provider_factory.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 // TestImage is a constant that specifies the default image used in all tests.
-const TestImage = "images:alpine/3.18/amd64"
+const TestImage = "ubuntu-minimal-daily:22.04"
```
MusicDin, line 14: Should we use `ubuntu-minimal` instead of `ubuntu-minimal-daily` for tests? Or is there any particular benefit of using daily builds?

MusicDin (changes requested):
Beside one question, LGTM.

MusicDin:
Also, can we ensure `ubuntu-minimal` is added to the LXD within actions for `5.0/stable` instead of disabling it?

simondeziel:
> Also, can we ensure `ubuntu-minimal` is added to the LXD within actions for `5.0/stable` instead of disabling it?

I thought of that and figured it'd be quicker to just disable it for now as we'll need to re-update those jobs once `5.0.3` hits `5.0/stable`.

internal/acctest/provider_factory.go
simondeziel: For those tests, it shouldn't matter much as we don't do anything inside the instances themselves, or very little.

I've been leaning toward `-daily` variants lately (in other test settings) because they better represent what users would run. The `-daily` one is essentially the same as the non-daily but with `apt-get update && apt-get upgrade` done inside of it.

MusicDin:
Okey, makes sense. Can you just rebase?

MusicDin (approved):

MusicDin (changes requested):
As per discussion, just needs to use different image for "lxd_cached_image" resources.

simondeziel:
@MusicDin this should address the race for cached image, please take another look (no rush).

MusicDin (approved):

MusicDin:
I like that you've added a separate variable for cached image. Thanks :)

PR #407, opened 2023-12-22
Wait exec output to be flushed and use -1 as default exec exit code

MusicDin:
Just a small improvement for exec:
- Wait for output to be flushed, so that we do not lose any output
- Use `-1` as default exec exit code, so that we can easily differentiate between command error and exec request error.

It also temporary disables the exit code check in one the exec tests. This is a regression in 5.19 and has been fixed in 5.20. As an alternative, we can use 5.18 client instead.

### DIFF
--- a/internal/common/lxd_exec.go
+++ b/internal/common/lxd_exec.go
@@ -62,34 +62,41 @@ func (e *ExecModel) Execute(ctx context.Context, server lxd.InstanceServer, inst
 		errBuf = utils.NewDiscardCloser()
 	}
 
-	execArgs := &lxd.InstanceExecArgs{
-		Stdout: outBuf,
-		Stderr: errBuf,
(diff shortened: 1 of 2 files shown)


simondeziel (approved):

tomponline:
@MusicDin @simondeziel can we add tests to the lxd test suite that test for the regression that occurred in lxd 5.19 to avoid it happening again?

simondeziel:
@tomponline we are planning to extend the LXD CI jobs on GH to pull the Terraform provider and run its testsuite there. I think that would have caught this but Din can confirm/infirm.

tomponline:
Mmm yeah let's discuss that, not 100% sure about that.

But we should have the lxd test suite itself be bolstered to test for things that this suite is testing for, rather than depend on yet another external dependency.

PR #413, opened 2024-01-12
Exec trigger and enabled attributes

MusicDin:
This PR modifies `exec` resource by replacing `triggers` (list of arbitrary string) with `trigger` and `enabled` attributes.

Attribute `trigger` has currently three possible values:
+ `on_change` (default) - Executes the command whenever the instance resource changes.
+ `on_start` - Executes the command on instance start.
+ `once` - Executes the command only once.
  
Setting attribute `enabled` to false, prevents the command from being executed.

In addition, list of exec blocks is converted into a map attribute (`execs`). If an exec is inserted at the beginning of the list, we cannot map blocks from the state with the blocks from the plan. This makes trigger `once` *impossible* to implement. In addition, order of execution is now defined alphabetically based on map keys.

We could still use exec blocks and add id attribute to them. IMO blocks look cleaner, but with maps is easier to reference specific exec (especially in case when order changes). Any thoughts?

### DIFF
--- a/docs/resources/instance.md
+++ b/docs/resources/instance.md
@@ -98,6 +98,8 @@ resource "lxd_instance" "container2" {
 
 * `file` - *Optional* - File to upload to the instance. See reference below.
 
+* `execs` - *Optional* - Map of exec commands to run within the instance. See reference below.
+
 * `limits` - *Optional* - Map of key/value pairs that define the
 	[instance resources limits](https://documentation.ubuntu.com/lxd/en/latest/reference/instance_options/#resource-limits).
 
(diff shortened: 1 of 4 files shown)


docs/resources/instance.md
```diff
@@ -141,12 +143,19 @@ The `file` block supports:
...
+  Defaults to `true`.
+
+* `trigger` - *Optional* - Determines when the command should be executed.
+  Available values are:
+  + `on_change` (default) - Executes the command whenever the instance resource changes.
+  + `on_start` - Executes the command on instance start. Note that the command will **not** be executed
+    if the instance is started outside of Terraform.
+  + `once` - Executes the command only once.
 
 * `environment` - *Optional* - Map of additional environment variables.
```
matifali, line 160: Are the environment variables injected into the container or are only available for the command?

MusicDin: Environment variables specified within the exec command are exclusive to that command and are not accessible elsewhere. To apply environment variables at the instance level, they must be defined in the `config`. 

For example:
```hcl
resource "lxd_instance" "inst" {
  name      = "c1"
  image     = "ubuntu-daily:22.04"

  config = {
    "environment.MY_ENV" = "value123"
  }

  execs = {
    "1" = {
      command       = ["sh", "-c", "echo $MY_ENV"]
      record_output = true
    }
  }
}

output "exec-output" {
  value = lxd_instance.inst.execs["1"].stdout # "value123\n"
}
```

LXD docs: https://documentation.ubuntu.com/lxd/en/latest/instance-exec/#environment

matifali: Great Thank you :)

simondeziel (approved):

PR #424, opened 2024-01-25
Generate unique names for test resources

MusicDin:
Occasionally the tests fail because the duplicate name gets generated for multiple test resources of the same type. This PR just wraps the `petname.Generate()` and generates a random string suffix if requested word count is more then 1.

### DIFF
--- a/internal/acctest/testutils.go
+++ b/internal/acctest/testutils.go
@@ -1,0 +1,28 @@
+package acctest
+
+import (
+	"math/rand"
+
+	petname "github.com/dustinkirkland/golang-petname"
+)
+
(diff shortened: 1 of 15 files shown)


simondeziel (commented):
Do you think it'd be enough to move from 2 to 3 words (`petname.Generate(3, "-")`) instead?

simondeziel:
`petname.Generate()` uses only 6 letter words by default it seems but we could bump that to 8-9.

MusicDin:
I would avoid relying on longer names, because in some cases it may be practical to test with shorter strings and we will run into same issue again. However, I agree that we should just keep it simple.

simondeziel (approved):

PR #426, opened 2024-01-30
github: remove DCO check replaced by DCO app

simondeziel:
This was tested by first pushing a commit without any `Signed-off-by` line in it. It properly flagged it as failing the DCO check.


simondeziel:
@MusicDin any idea why `5.0/edge` is broken?

MusicDin:
From the logs, it seems that the instance is having issues being stopped, but I'm not sure why. I've tried quickly reproducing it locally, but without any success. Will have to take a closer look at this.

MusicDin:
Seems that the issue is not consistent, which will be a bit tricky to investigate. I've filled an issue for this (#428), so we can get this PR merged.

MusicDin (approved):

PR #427, opened 2024-02-01
expose new ipv4_addresses and ipv6_addresses computed attributes on i…

### LINKED ISSUE #321: Support multiple ip addresses on single interface
We have a setup where two ipv6 subnets are used on single interface, example:

```
+----------------------+---------+---------------------+---------------------------------------------+-----------+-----------+--------------------+
| vault0               | RUNNING | 10.30.10.49 (eth0)  | fd19:cafe:510:0:216:3eff:fef3:6194 (eth0)   | CONTAINER | 0         | node1              |
|                      |         |                     | 3ac7:ffa:841:7d19:116:4eff:5ef3:1194 (eth0) |           |           |                    |
+----------------------+---------+---------------------+---------------------------------------------+-----------+-----------+--------------------+
```

`lxd_container` output will return only one ipv6 address, the public one; the private one won't be returned.

Is it acceptable to convert `ip_address`, ` ipv4_address` and `ipv6_address` output type to list so that multiple addresses can be returned ? To preserve compatibility this field can stay as is if only one address is detected.

aanderse:
…nstance resources

resolves #321 - see motivation and reasoning in linked issue... but basically multiple ip addresses assigned to a single interface is something the current provider does not cover

how it looks when 2 ipv6 addresses are assigned to a single interface:
```
aanderse@obsidian ~> jq '.lxd_instance.dash' resources.json
{
  "config": {
    "boot.autostart": "true"
  },
  "description": "",
  "device": [],
  "ephemeral": false,
  "execs": null,
  "file": [],
  "image": "nixos",
  "ipv4_address": "10.30.10.227",
  "ipv6_address": "2a01:4f8:242:5e07:216:3eff:fe67:ba0f",
  "ipv6_addresses": [
    "fd23:cafe:618:0:216:3eff:fe67:ba0f",
    "2a01:4f8:242:5e07:216:3eff:fe67:ba0f"
  ],
  "limits": {},
  "mac_address": "00:16:3e:67:ba:0f",
  "name": "dash",
  "profiles": [
    "default"
  ],
  "project": "sky",
  "remote": null,
  "running": true,
  "status": "Running",
  "target": "618-fsn1-dc15-hetz",
  "timeouts": null,
  "type": "container",
  "wait_for_network": true
}
```

i've never written or contributed to a `terraform` provider before so please let me know if i've completely missed anything obvious here...

thank you for your time and any potential reviews

### DIFF
--- a/internal/instance/resource_instance.go
+++ b/internal/instance/resource_instance.go
@@ -54,7 +54,9 @@ type InstanceModel struct {
 
 	// Computed.
 	IPv4   types.String `tfsdk:"ipv4_address"`
+	IPv4s  types.List   `tfsdk:"ipv4_addresses"`
 	IPv6   types.String `tfsdk:"ipv6_address"`
+	IPv6s  types.List   `tfsdk:"ipv6_addresses"`
 	MAC    types.String `tfsdk:"mac_address"`
 	Status types.String `tfsdk:"status"`
 
@@ -315,8 +317,18 @@ func (r InstanceResource) Schema(ctx context.Context, _ resource.SchemaRequest, 
 				Computed: true,
 			},
 
+			"ipv4_addresses": schema.ListAttribute{
+				Computed:    true,
+				ElementType: types.StringType,
+			},
+
 			"ipv6_address": schema.StringAttribute{
 				Computed: true,
+			},
+
+			"ipv6_addresses": schema.ListAttribute{
+				Computed:    true,
+				ElementType: types.StringType,
 			},
 
 			"mac_address": schema.StringAttribute{
@@ -1015,7 +1027,9 @@ func (r InstanceResource) SyncState(ctx context.Context, tfState *tfsdk.State, s
 	// Reset IPv4, IPv6, and MAC addresses. If case instance has lost
 	// network connectivity, we should reflect that in state.
 	m.IPv4 = types.StringNull()
+	m.IPv4s = types.ListNull(types.StringType)
 	m.IPv6 = types.StringNull()
+	m.IPv6s = types.ListNull(types.StringType)
 	m.MAC = types.StringNull()
 
 	// First there is an access_interface set, extract IPv4, IPv4, and
@@ -1032,9 +1046,19 @@ func (r InstanceResource) SyncState(ctx context.Context, tfState *tfsdk.State, s
 			accIfaceFound = true
 		}
 
+		ipv4s, ok := findIPv4Addresses(net)
+		if ok {
+			m.IPv4s, _ = types.ListValueFrom(ctx, types.StringType, ipv4s)
+		}
+
 		ipv6, ok := findIPv6Address(net)
 		if ok {
 			m.IPv6 = types.StringValue(ipv6)
+		}
+
+		ipv6s, ok := findIPv6Addresses(net)
+		if ok {
+			m.IPv6s, _ = types.ListValueFrom(ctx, types.StringType, ipv6s)
 		}
 	}
 
@@ -1052,9 +1076,19 @@ func (r InstanceResource) SyncState(ctx context.Context, tfState *tfsdk.State, s
 				m.MAC = types.StringValue(mac)
 			}
 
+			ipv4s, ok := findIPv4Addresses(net)
+			if ok {
+				m.IPv4s, _ = types.ListValueFrom(ctx, types.StringType, ipv4s)
+			}
+
 			ipv6, ok := findIPv6Address(net)
 			if ok {
 				m.IPv6 = types.StringValue(ipv6)
+			}
+
+			ipv6s, ok := findIPv6Addresses(net)
+			if ok {
+				m.IPv6s, _ = types.ListValueFrom(ctx, types.StringType, ipv6s)
 			}
 		}
 	}
@@ -1335,6 +1369,18 @@ func findIPv4Address(network api.InstanceStateNetwork) (string, string, bool) {
 	return ipv4, mac, (ipv4 != "")
 }
 
+// findIPv4Addresses searches the network for any IPv4 addresses.
+func findIPv4Addresses(network api.InstanceStateNetwork) ([]string, bool) {
+	var ipv4 []string
+	for _, ip := range network.Addresses {
+		if ip.Family == "inet" {
+			ipv4 = append(ipv4, ip.Address)
+		}
+	}
+
+	return ipv4, (len(ipv4) != 0)
+}
+
 // Find last global IPv6 address or return any last IPv6 address
 // if there is no global address. This works analog to the IPv4
 // selection mechanism but favors global addresses.
@@ -1358,4 +1404,29 @@ func findIPv6Address(network api.InstanceStateNetwork) (string, bool) {
 	}
 
 	return ipv6, (ipv6 != "")
+}
+
+// Find all global IPv6 addresses or return any IPv6 address
+// if there are no global addresses. This works analog to the IPv4
+// selection mechanism but favours global addresses.
+func findIPv6Addresses(network api.InstanceStateNetwork) ([]string, bool) {
+	var ipv6 []string
+
+	for _, ip := range network.Addresses {
+		if ip.Family == "inet6" && ip.Scope == "global" {
+			ipv6 = append(ipv6, ip.Address)
+		}
+	}
+
+	if len(ipv6) != 0 {
+		return ipv6, true
+	}
+
+	for _, ip := range network.Addresses {
+		if ip.Family == "inet6" {
+			ipv6 = append(ipv6, ip.Address)
+		}
+	}
+
+	return ipv6, (len(ipv6) != 0)
 }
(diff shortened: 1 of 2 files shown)


internal/instance/resource_instance.go
```diff
@@ -1359,3 +1405,29 @@ func findIPv6Address(network api.InstanceStateNetwork) (string, bool) {
...
+// Find all global IPv6 addresses or return any IPv6 address
+// if there are no global addresses. This works analog to the IPv4
+// selection mechanism but favours global addresses.
+func findIPv6Addresses(network api.InstanceStateNetwork) ([]string, bool) {
+	var ipv6 []string
+
+	for _, ip := range network.Addresses {
+		if ip.Family == "inet6" && ip.Scope == "global" {
+			ipv6 = append(ipv6, ip.Address)
+			// ipv6 = ip.Address
```
simondeziel, line 1418: Left over comment?

aanderse: and this is why i'm not supposed to make PRs late at night :sweat_smile:

thanks

simondeziel:
@aanderse Thanks, it LGTM but could you please add the `Signed-off-by` to your commit? Clicking on the DCO check details will walk you through how to do that.

MusicDin:
Thanks for you contribution.

Could you please add documentation for new exported attributes?

In addition, I think we should add some tests for this to ensure the desired IP addresses are listed. Speaking of that, I was expecting the address lists to include *all* system IP addresses not just from a specific interface. @simondeziel is this the way we want to proceed?

aanderse:
> Could you please add documentation for new exported attributes?

definitely! i think latest push is good, but let me know if want something else

> In addition, I think we should add some tests for this to ensure the desired IP addresses are listed.

unfortunately i don't think i can do this :scream: i just spent a few hours trying to run the existing basic tests over 3 different systems and i get a variety of errors, depending on the system i ran on :disappointed:
without being able to run the basic tests i have no hope of adding a new one

given how simple an additional test would be to add for someone who is familiar with the system and _can_ run tests... is anyone interested in helping me out? :bowing_man:

> Speaking of that, I was expecting the address lists to include _all_ system IP addresses not just from a specific interface.

this PR **does** accomplish what you are requesting :heavy_check_mark: specifically, this PR follows the existing behaviour in that all IP addresses for all interfaces are returned, unless `user.access_interface` is specified in which case only the IP address(es) for that interface will be returned

---

please let me know how we can proceed and thanks for your review!

MusicDin:
> definitely! i think latest push is good, but let me know if want something else

Thanks. Seems to be good.

> i just spent a few hours trying to run the existing basic tests over 3 different systems and i get a variety of errors, depending on the system i ran on 

Could you maybe provide us with some reproducers, since the tests are expected to work on *most* LXD versions (some test may need to be constrained for older LXD versions, if those do not implement a specific feature). Also, which LXD/Terraform version are you using?

> is anyone interested in helping me out?

Yep, absolutely, I will help you.

> this PR follows the existing behaviour in that all IP addresses for all interfaces are returned, unless user.access_interface is specified in which case only the IP address(es) for that interface will be returned

In case the `user.access_interface` is not set the list of `ipv*_addresses` will contain only the addresses from the last interface that contains at least one address of the searched IP family.

A quick demo:
```hcl
resource "lxd_network" "internal" {
  name = "internal"
}

resource "lxd_instance" "inst" {
  name  = "test"
  image = "ubuntu:22.04"

  device {
    name = "eth0"
    type = "nic"
    properties = {
      network = lxd_network.internal.name
    }
  }

  device {
    name = "eth1"
    type = "nic"
    properties = {
      network = "lxdbr0"
    }
  }
}

output "inst-test" {
  value = {
    "ipv4s" = lxd_instance.inst.ipv4_addresses
    "ipv6s" = lxd_instance.inst.ipv6_addresses
  }
}
```

Output:
```hcl
inst-test = {
  "ipv4s" = tolist([
    "192.168.255.61",
  ])
  "ipv6s" = tolist([
    "fd42:e8b9:6cf1:52c4:216:3eff:fe8b:ff78",
  ])
}
```

And lxc output:
```
$ lxc list test -c n46
+------+-----------------------+-----------------------------------------------+
| NAME |         IPV4          |                     IPV6                      |
+------+-----------------------+-----------------------------------------------+
| test | 192.168.255.61 (eth0) | fd42:e8b9:6cf1:52c4:216:3eff:fe8b:ff78 (eth1) |
|      |                       | fd42:474b:622d:259d:216:3eff:fedb:c80a (eth0) |
+------+-----------------------+-----------------------------------------------+
```

The first issue is that private IPv6 are not present in the list. This is because the `findIPv6Addresses` first searches only for global addresses. 

The second issue is that it returns only the IP addresses from the last interface.
While this is documented for `ipv4_address` and `ipv6_address` in [instance network section](https://registry.terraform.io/providers/terraform-lxd/lxd/latest/docs/resources/instance#instance-network-access), it would probably be less confusing if the returned IPv6 would be from the same interface on which IPv4 is found.

The above issues can be easily fixed for this PR. 

However, I was discussing with @simondeziel that it may be more beneficial if the exported attribute contains IP addresses mapped with the interface name. This way it would also allow to extract only the addresses for a specific interface. With the current approach this would be possible for a single interface using `user.access_interface`, but not if two IPs from different interfaces are required. 

Any thoughts about that? Also, so you maybe have a specific use case that could help us narrow down the requirements?

aanderse:
> Could you maybe provide us with some reproducers, since the tests are expected to work on _most_ LXD versions (some test may need to be constrained for older LXD versions, if those do not implement a specific feature). Also, which LXD/Terraform version are you using?

```console
aaron@yokozuna ~/w/terraform-provider-lxd (main)> lxc --version
5.19
aaron@yokozuna ~/w/terraform-provider-lxd (main)> terraform --version
Terraform v1.7.1-dev
on linux_amd64

Your version of Terraform is out of date! The latest version
is 1.7.2. You can update by downloading from https://www.terraform.io/downloads.html
```

i was consistently getting different network errors across different machines, some related to `dnsmasq` like this:
```console
aaron@yokozuna ~/w/terraform-provider-lxd (main)> env TF_ACC=1 go test -v -timeout 30s -run '^TestAccInstance_accessInterface$' github.com/terraform-lxd/terraform-provider-lxd/internal/instance
=== RUN   TestAccInstance_accessInterface
    resource_instance_test.go:1005: Step 1/1 error: Error running apply: exit status 1
        
        Error: Failed to create network "gelding-VTcebd"
        
          with lxd_network.network1,
          on terraform_plugin_test.tf line 12, in resource "lxd_network" "network1":
          12: resource "lxd_network" "network1" {
        
        The DNS and DHCP service exited prematurely: exit status 2 ("dnsmasq: failed
        to create listening socket for 10.150.19.1: Address already in use")
--- FAIL: TestAccInstance_accessInterface (0.30s)
FAIL
FAIL    github.com/terraform-lxd/terraform-provider-lxd/internal/instance       0.302s
FAIL
```

> In case the `user.access_interface` is not set the list of `ipv*_addresses` will contain only the addresses from the last interface that contains at least one address of the searched IP family.

ah yeah, i missed that one loop... understood

> However, I was discussing with @simondeziel that it may be more beneficial if the exported attribute contains IP addresses mapped with the interface name. This way it would also allow to extract only the addresses for a specific interface. With the current approach this would be possible for a single interface using `user.access_interface`, but not if two IPs from different interfaces are required.
> 
> Any thoughts about that? Also, so you maybe have a specific use case that could help us narrow down the requirements?

i'm happy with any direction you want to go as my requirements are fairly straight forward: i have multiple IP addresses assigned to a single interface and need access to them from something like `terraform output`
as you proposed - a map returning a list of addresses per interface - sounds more generally useful to a larger audience :+1:

thanks again for your time

MusicDin:
We can follow the existing LXD network interface structure when exporting:
```
interfaces = []Interface{
  {
    name: eth0
    type: broadcast
    mac: ""
    state: ""
    ip_addresses: [
      ...
    ]
  }
}
```

But the above approach may be a bit clunky when parsing values from HCL.
Therefore, maybe we could do something like this, to flatten the resulting structure as much as possible?
```
addresses = []IP{
  {
    address: ".."
    family: "v4/v6"
    scope: "local/gobal/link"
    interface: "eht0"
  }
}
```

aanderse:
@MusicDin either of the above approaches works well for my purposes... in my case the data ends up parsed as `json` for me so i'm not picky

i'm very interested in getting this functionality added to the provider so please let how you want this implemented and hopefully my `go`+`terraform` sdk knowledge will be sufficient enough for the task at hand

thanks again :bowing_man:

MusicDin:
Let's proceed with the second approach then:

```
addresses = []IP{
  {
    address: ".."
    family: "v4/v6"
    scope: "local/gobal/link"
    interface: "eht0"
  }
}
```

And thank you very much for contributing 👍

aanderse:
sorry for dropping the ball on this but it fell on priority list... thanks so much!

PR #460, opened 2024-05-16
Export instance's network interfaces

### LINKED ISSUE #321: Support multiple ip addresses on single interface

### LINKED ISSUE #453: ipv4_address only exposes a single interface, tends to change to an unreachable address if one is configured
Currently lxd_instance.name.ipv4_address only exposes a single address/interface. If you have multiple interfaces configured, the value may switch to an different or private/inaccessible IP address.

There is currently no way to pick one of the others, even manually.

### Specific use case
I am using terraform to deploy LXD VMs, exporting the instances as ansible inventory and then using ansible to configure them. As part of that, I install libvirt-daemon to deploy VMs inside the VM (in this case to test a libvirt issue).

Once you install libvirt-daemon, it creates and sets up a virtual bridge (virbr0) with the IP 192.168.122.1 (which is not routable from outside the VM).

Next time I re-run terraform it always switches the ipv4_address from the previously good public IP (enp5s0/172.16.0.*) to the private IP (virbr0/192.168.122.1).

This means ansible can no longer connect and configure the instance, even though it worked at the start.

It is notable that in my case I am using an unmanaged network for the instance, with a 'type: bridged' NIC and LXD is not performing IPAM. I am also currently using VMs. I have not tested the behaviour with containers or when LXD is doing IPAM. The IPs are reported back by the LXD agent inside the VM.

### Proposed Resolution
Ideally, LXD would know which interface had the default route, and pick that as a preference.

It doesn't currently seem to expose that in the API, ideally it could, but in the absence of that, exposing all of the IP addresses would at least allow someone to pick one by interface name. It's not ideal as that interface name can change depending on the workload, but would at least give a workable option for this use case.

`lxc ls` exposes the full list:
```
$ lxc ls
+------+---------+------------------------+------+-----------------+-----------+
| NAME |  STATE  |          IPV4          | IPV6 |      TYPE       | SNAPSHOTS |
+------+---------+------------------------+------+-----------------+-----------+
| vm1  | RUNNING | 192.168.122.1 (virbr0) |      | VIRTUAL-MACHINE | 0         |
|      |         | 172.16.0.206 (enp5s0)  |      |                 |           |
+------+---------+------------------------+------+-----------------+-----------+
| vm2  | RUNNING | 193.1.1.1 (z-test)     |      | VIRTUAL-MACHINE | 0         |
|      |         | 192.168.122.1 (virbr0) |      |                 |           |
|      |         | 173.168.200.1 (a-test) |      |                 |           |
|      |         | 172.16.0.207 (enp5s0)  |      |                 |           |
+------+---------+------------------------+------+-----------------+-----------+
```

`lxc ls` seems to be ordering the list alphabetically by IP address, so the order is predictable but which interface you get will be dependant on what subnet and IP address you are using for the different interfaces.

terraform seems to be using a different order, I did not attempt to figure out what it was. But it wasn't obviously ordered by the IP or name, possibly by the most recently added interface.

MusicDin:
Export map of network interfaces as an instance attribute. The map key represents the network device name (from LXD configuration).

In addition, attributes `ipv4_address` and `ipv6_address` are now exported from the same interface and only global IPs are exported. The IPs are retrieved from the first network interface (sorted alphabetically) that contains at least one global IP.

Fixes #453
Fixes #321
Closes #427

### DIFF
--- a/internal/common/lxd_interface.go
+++ b/internal/common/lxd_interface.go
@@ -1,0 +1,111 @@
+package common
+
+import (
+	"context"
+	"strings"
+
+	"github.com/canonical/lxd/shared/api"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// Interface represents LXD instance network interface.
+type InterfaceModel struct {
+	// Real name of the interface within the instance. If config interface is
+	// defined as "eth0", the real interface within a container will have the
+	// same name. However, VMs will most likely have enp0sX.
+	RealName types.String `tfsdk:"name"`
+
+	// State of the interface [UP, DOWN].
+	State types.String `tfsdk:"state"`
+
+	// Type of the interface [broadcast, loopback].
+	Type types.String `tfsdk:"type"`
+
+	// List of interface addresses.
+	Addresses types.List `tfsdk:"ips"`
+}
+
+// Interface IP is a wrapper of the interface IP address.
+type IPModel struct {
+	// IP address.
+	Address types.String `tfsdk:"address"`
+
+	// Family [inet, inet6].
+	Family types.String `tfsdk:"family"`
+
+	// Scope [link, local, global].
+	Scope types.String `tfsdk:"scope"`
+}
+
+// ToInterfaceMapType converts provided intance networks into types.Map. The function
+// also accepts instance config which is used to determine configuration interface name.
+func ToInterfaceMapType(ctx context.Context, instNetworks map[string]api.InstanceStateNetwork, instConfig map[string]string) (types.Map, diag.Diagnostics) {
+	ipType := map[string]attr.Type{
+		"address": types.StringType,
+		"family":  types.StringType,
+		"scope":   types.StringType,
+	}
+
+	infType := map[string]attr.Type{
+		"name":  types.StringType,
+		"state": types.StringType,
+		"type":  types.StringType,
+		"ips":   types.ListType{ElemType: types.ObjectType{AttrTypes: ipType}},
+	}
+
+	nilMap := types.MapNull(types.ObjectType{AttrTypes: infType})
+	if len(instNetworks) == 0 {
+		return nilMap, nil
+	}
+
+	interfaces := make(map[string]InterfaceModel, len(instNetworks))
+	for name, net := range instNetworks {
+		// Find volatile entry that contains mac address of the network
+		// interface. If addresses match, extract the config name of the
+		// interface from the config key (volatile.<if_name>.hwaddr).
+		cfgInfName := ""
+		for k, v := range instConfig {
+			if v == net.Hwaddr {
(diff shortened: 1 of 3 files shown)


MusicDin:
Adding a quick example:
```hcl
resource "lxd_network" "internal" {
  name      = "internal"
}

resource "lxd_instance" "instance" {
  name      = "inst"
  image     = "ubuntu-daily:22.04"
  type      = "virtual-machine"

  device {
    name = "eth1"
    type = "nic"

    properties = {
      nictype = "bridged"
      parent  = lxd_network.internal.name
    }
  }
}

output "output" {
  value = {
    "interfaces": lxd_instance.instance.interfaces
  }
}
```

An example output (`eth0` from default profile and `eth1` from custom network):
```
output = {
  "interfaces" = tomap({
    "eth0" = {
      "ips" = tolist([
        {
          "address" = "10.191.107.100"
          "family" = "inet"
          "scope" = "global"
        },
        {
          "address" = "fd42:e389:6d60:1521:216:3eff:fe55:7972"
          "family" = "inet6"
          "scope" = "global"
        },
        {
          "address" = "fe80::216:3eff:fe55:7972"
          "family" = "inet6"
          "scope" = "link"
        },
      ])
      "name" = "enp5s0"
      "state" = "up"
      "type" = "broadcast"
    }
    "eth1" = {
      "ips" = tolist([])
      "name" = "enp6s0"
      "state" = "down"
      "type" = "broadcast"
    }
  })
}
```

internal/common/lxd_interface.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// Interface represents LXD instance network interface.
+type InterfaceModel struct {
+	// Real name of the interface within the instance. If config interface is
+	// defined as "eth0", the real interface within a container will have the
+	// same name. However, VMs will most likely have enp0sX.
```
simondeziel, line 17: For VMs, it's usually `enpXs0` where X starts at 5.

simondeziel (approved):

internal/common/lxd_interface.go
MusicDin: Thanks, will fix :)

PR #469, opened 2024-05-30
Allow remote LXD authentication using token

MusicDin:
Currently, provider supports only initial authentication using trust password. This PR adds support for `token` authentication. Attributes `token` and `password` are mutually exclusive.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -46,10 +46,10 @@ provider "lxd" {
   }
 
   remote {
-    name     = "lxd-server-2"
-    scheme   = "https"
-    address  = "10.1.2.8"
-    password = "password"
+    name    = "lxd-server-2"
+    scheme  = "https"
+    address = "10.1.2.8"
+    token   = "token"
   }
 }
 ```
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -90,7 +90,12 @@ The `remote` block supports:
 
 * `name` - *Optional* - The name of the LXD remote.
 
-* `password` - *Optional* - The password to authenticate to the LXD remote.
+* `password` - *Optional* - The [trust password](https://documentation.ubuntu.com/lxd/en/latest/authentication/#adding-client-certificates-using-a-trust-password)
+  to authenticate to the LXD remote. This method is **not recommended** and will most likely be
+  removed from LXD in the future releases. Use `token` instead.
```
simondeziel, line 95:
```suggestion
  removed from LXD in a future release. Use `token` instead.
```

docs/index.md
```diff
@@ -90,7 +90,12 @@ The `remote` block supports:
...
+
+* `token` - *Optional* - The [token](https://documentation.ubuntu.com/lxd/en/latest/authentication/#adding-client-certificates-using-tokens)
+  to authenticate to the LXD remote.
```
simondeziel, line 98: I wouldn't say it is used to authenticate to the LXD remote. It's more to securely add the LXD remote locally (due to the embedded fingerprint) and also to inject the client certificate into LXD's trust store.

internal/provider/provider.go
```diff
@@ -118,7 +120,17 @@ func (p *LxdProvider) Schema(_ context.Context, _ provider.SchemaRequest, resp *
...
 							Optional:    true,
 							Sensitive:   true,
-							Description: "The password for the remote.",
+							Description: "The trust password to authenticate to the LXD remote.",
+						},
+
+						"token": schema.StringAttribute{
+							Optional:    true,
+							Sensitive:   true,
+							Description: "The token to authenticate to the LXD remote",
```
simondeziel, line 129: One last nit: please update that description with the one from doc.

MusicDin: Is the description in docs accurate enough, or would you like me to further expand it?
As now there is a link to the documentation, I assumed it is enough to emphasize that password/token is used for initial authentication.

simondeziel: The description in the doc is perfect, I'd probably reuse it as-is but strip the link/markdown syntax.

simondeziel (approved):

PR #471, opened 2024-05-31
Add support for client certificates

### LINKED ISSUE #408: support for adding client certificates
It would be nice if it were possible to add client certificates to a LXD host through the Terraform provider.

The equivalent of this on the CLI:

```
lxc config trust add - <<EOF
<certificate>
EOF
```
source: https://documentation.ubuntu.com/lxd/en/latest/authentication/#adding-trusted-certificates-to-the-server

MusicDin:
Add support for adding client certificates into LXD trust store.

Example:
```hcl
resource "lxd_trust_certificate" "cert" {
  name     = "my-cert"
  type     = "client"         # client or metrics
  path     = "/path/to/cert"
  content = "..."             # Mutually exclusive with "path"
  projects = []               # Projects to restrict the certificate to. Certificate is restricted if len(projects) > 1
  remote   = ""
}
```

Computed field is `fingerprint`.

---

TODO: 
- [x] Agree on naming of the resource - I just went with `trust_certificate`, but this should be changed.
- [x] Docs: Once we agree on naming.

---

Fixes #408

### DIFF
--- a/docs/resources/trust_certificate.md
+++ b/docs/resources/trust_certificate.md
@@ -1,0 +1,46 @@
+# lxd_trust_certificate
+
+The `lxd_trust_certificate` resource allows you to register new client certificates in the LXD trust store.
+
+## Example Usage
+
+```hcl
+resource "lxd_trust_certificate" "cert1" {
+  name = "cert1"
+  path = "/path/to/cert"
+}
+
+resource "lxd_trust_certificate" "cert2" {
+  name    = "cert2"
+  content = <<EOF
+-----BEGIN CERTIFICATE-----
+...
+-----END CERTIFICATE-----
+EOF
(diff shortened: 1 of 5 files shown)


simondeziel (approved):
LGTM, thanks

MusicDin:
@simondeziel Shall we keep `trust_certificate` or do you have any better naming?

simondeziel (changes requested):
The name `trust_certificate` is OK to me. However, please add support for a certificate type as it's not always just `client` certs, we have `metrics` ones too.

simondeziel (commented):
Do you think `type=metrics` would warrant a test case?

simondeziel (approved):

PR #475, opened 2024-06-03
Use Alpine images for cached image test

MusicDin:
Use smaller Alpine images for testing cached images.


MusicDin:
I'm not sure why, but using `ubuntu` remote it occasionally takes >5min to copy an image.

For example https://github.com/terraform-lxd/terraform-provider-lxd/actions/runs/9345944914 took 32 minutes

PR #481, opened 2024-06-10
goreleaser: re-enable arm64 builds for Windows

simondeziel:
I also considered dropping the `arm32` builds for Windows but it seems MS still supports an OS for that CPU arch: https://en.wikipedia.org/wiki/ARM_architecture_family#32-bit_operating_systems


MusicDin (approved):
Thanks 🚀

PR #488, opened 2024-06-27
docs/resources/instance: add multiple ordered `execs`

### LINKED ISSUE #486: Question: what is the order of commands execution in the "execs" block?
Could you please clarify the order of commands execution in the `execs` block in the `lxd_instance` resource? Is it:
- order elements defined in `execs` block?
- alphabetic order?
- undetermined?

simondeziel:
Fixes #486


MusicDin (approved):
Thanks 🚀

PR #490, opened 2024-07-03
Fix cached image being searched only in default project

MusicDin:
When image remote is not set, use instance server as an image server.

Previously, when the image server (an actual LXD server) was retrieved, the project was not set. This resulted in images always being searched only in the default project. Now, if image remote is not provided, we use the instance server where the instance is being created as an image server.

Example:
```hcl
resource "lxd_project" "proj" {
  name = "proj"
  config = {
    "features.images"   = true
    "features.profiles" = false
  }
}

resource "lxd_cached_image" "alpine" {
    source_remote = "images"
    source_image  = "alpine/3.18"
    project       = lxd_project.proj.name
}

resource "lxd_instance" "c1" {
    name    = "c1"
    image   = lxd_cached_image.alpine.fingerprint
    project = lxd_project.proj.name
    running = false
}
```

---

Whenever an image remote is set (for example to `local`) the image will be searched in `default` project. 

```hcl
resource "lxd_cached_image" "alpine" {
    source_remote = "images"
    source_image  = "alpine/3.18"
    project       = lxd_project.proj.name
}

resource "lxd_instance" "c1" {
    name    = "c1"
    image   = "local:${lxd_cached_image.alpine.fingerprint}" # Image not found
    project = lxd_project.proj.name
    running = false
}
```

### DIFF
--- a/internal/image/resource_cached_image_test.go
+++ b/internal/image/resource_cached_image_test.go
@@ -210,6 +210,29 @@ func TestAccCachedImage_project(t *testing.T) {
 	})
 }
 
+func TestAccCachedImage_instanceFromImageFingerprint(t *testing.T) {
+	projectName := acctest.GenerateName(2, "")
+	instanceName := acctest.GenerateName(2, "")
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctest.PreCheck(t) },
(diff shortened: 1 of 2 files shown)


simondeziel (approved):
LGTM, thanks!

PR #492, opened 2024-07-05
Support null values in config - from Incus

MusicDin:
Cherry picked from https://github.com/lxc/terraform-provider-incus/pull/82


simondeziel (changes requested):
The commit 1st message would need a `s/incus/lxd/g`, please.

PR #493, opened 2024-07-05
Add network forward and ACL resources

MusicDin:
Cherry-picked from https://github.com/lxc/terraform-provider-incus/pull/84 and https://github.com/lxc/terraform-provider-incus/pull/87

Needs rebase after #492


internal/network/resource_network_forward.go
```diff
@@ -0,0 +1,466 @@
...
+		ports, diags := ToNetworkForwardPortList(ctx, config.Ports)
+		if diags.HasError() {
+			resp.Diagnostics.Append(diags...)
+			return
+		}
+
+		for i := range ports {
+			if ports[i].TargetPort == "" {
+				ports[i].TargetPort = ports[i].ListenPort
+			}
```
simondeziel, line 201: Why not leave them unspecified? This way, LXD is the bit controlling what default gets applied.

MusicDin: To show the user the resulting value in the Terraform plan and to ensure the value is always applied even if manually changed in LXD. We can also leave it unset.

simondeziel: I'd prefer to leave it unset so that LXD's authoritative in the default behavior.

simondeziel (approved):

PR #494, opened 2024-07-05
Fix network zone minor issue - from Incus

MusicDin:
Cherry picked from https://github.com/lxc/terraform-provider-incus/pull/81

### DIFF
--- a/docs/resources/network_zone_record.md
+++ b/docs/resources/network_zone_record.md
@@ -22,7 +22,7 @@ resource "lxd_network_zone" "zone" {
 
 resource "lxd_network_zone_record" "record" {
   name = "ns"
-  zone = lxd_network_zone.zone.id
+  zone = lxd_network_zone.zone.name
 
   entry {
       type  = "CNAME"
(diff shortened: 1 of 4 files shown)


docs/resources/network_zone_record.md
```diff
@@ -62,7 +62,7 @@ The `entry` block supports:
 
 * `value` - **Required** - Entry value.
 
-* `ttl` - *Optional* - Entry time to live (TTL).
+* `ttl` - *Optional* - Entry time to live (TTL) in seconds. Defaults to 300 seconds.
```
simondeziel, line 65: If not specified, a record's TTL would normally come from the zone's default TTL. As such, I think it'd be best to not provide a default value.

MusicDin: Thanks for pointing that out. I can fix that, however, I was following the docs: https://documentation.ubuntu.com/lxd/en/latest/howto/network_zones/#add-or-remove-entries 

> You can use the --ttl flag to set a custom time-to-live (in seconds) for the entry. Otherwise, the default of 300 seconds is used.

How is the network zone default TTL defined?

simondeziel: LXD doesn't seem to set a zone default but a record default of 300: https://github.com/canonical/lxd/blob/main/lxd/network/zone/record.go#L206-L208

simondeziel (approved):

PR #510, opened 2024-08-07
Allow managing default profiles in non-default projects

MusicDin:
Allow managing default profile in non-default projects. When `default` profile is created, instead of trying to create it, fetch the existing, update it, and inject it into the Terraform state.

Default profile cannot be managed if it is located in project `default` or if the project has `features.profiles` set to `false` (disabled).

### DIFF
--- a/internal/profile/resource_profile_test.go
+++ b/internal/profile/resource_profile_test.go
@@ -2,6 +2,7 @@ package profile_test
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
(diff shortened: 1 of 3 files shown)


internal/profile/resource_profile_test.go
```diff
@@ -233,6 +234,37 @@ func TestAccProfile_project(t *testing.T) {
...
+	projectName := acctest.GenerateName(2, "-")
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctest.PreCheck(t) },
+		ProtoV6ProviderFactories: acctest.ProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				// Ensure default profile cannot be managed in default project.
+				Config:      testAccProfile_basic("default"),
+				ExpectError: regexp.MustCompile("Cannot import existing profile"),
```
simondeziel, line 247: Is a regexp actually needed?

internal/profile/resource_profile_test.go
```diff
@@ -233,6 +234,37 @@ func TestAccProfile_project(t *testing.T) {
...
+			},
+			{
+				// Ensure default profile cannot be managed in project with "features.profiles".
```
simondeziel, line 250: Please add a corresponding comment to the "default profile can be managed in non-default projects" case.

simondeziel (approved):
LGTM with 2 small nits.

internal/profile/resource_profile_test.go
MusicDin: Yes, this is the expected type of `ExpectError` field.

MusicDin: Just attaching here:
```
field ExpectError *regexp.Regexp

ExpectError allows the construction of test cases that we expect to fail with an
error. The specified regexp must match against the error for the test to pass.

This functionality is only intended for provider-controlled error messaging. 
While in certain scenarios this can also catch testing logic error messages, 
those messages are not protected by compatibility promises.
```

simondeziel (approved):

Issue #487, opened 2024-06-27
Add ability to generate config trust tokens


stew3254:
I am currently working on a scenario where I deploy a MAAS container and would like to connect it to the host LXD in order to spin up VMs to use within MAAS. This is a nice system because it can fully utilize the host's resources, and can be fully automatic by using a LXD trust password. However, the [trust password is going away](https://github.com/canonical/lxd/pull/13567) and there is no way for me to generate a token via Terraform to supply to my container so it can add its client cert.

If possible, I propose a new resource be created which allows users to create trust tokens and a method by which they can be consumed by another resource.

MusicDin:
Hmm, I see where this is going and I like the idea, but the issue with `trust_token` resource is that it is a one time operation, which we want to avoid as much as possible. Ideally the resource configuration should be reflected in the LXD, which is not the case for tokens (as they are required only once). 

While the creation of a token can be easily implemented, my question is what do we do with the resource (and its state) once the token is generated (can we avoid having it in a state) and consumed (do we just remove it and ignore the resource on future applys)? 

@simondeziel any thoughts?

---

Another potential approach would be to have a function that would request a token with a given name when called. If the token already exists, it would return it, otherwise it would request a new one. This way, we don't have to track the state for this resource, as all the user cares is to get the token *that works*. The downside would be that we cannot clean it up if it is not consumed. 

Just to emphasize, I'm not sure if this would work - just an idea.

---

Finally, could the new `trust_certificate` be of use in this scenario? It would require you to provide an existing client certificate. Therefore, it is not ideal, but may serve as a temporary workaround. (this will be included in next terraform release - [unreleased docs](https://github.com/terraform-lxd/terraform-provider-lxd/blob/main/docs/resources/trust_certificate.md))

stew3254:
Hello @MusicDin. I realized there might be some complication in both of the first two scenarios,  but I'm still new to Terraform and learning the model. I'm not sure I can provide much input on the two, but am happy to learn!

Regarding the third, it might work in my case. I'm using Terraform to generate the MAAS machine, so I'm not sure I can force it to have a specific client certificate. It might be possible for me to work around by using a bind mount device to provide the cert and and cloud-init script to replace whatever LXD generated. Then, Terraform would also be able to see the file and add it to the local trust store.

simondeziel:
As explained by Din, the ephemeral nature of the remote join token that it's requested and consumed only once makes it a bad fit for automation via Terraform.

[...]

PR #512, opened 2024-08-07
Add trust token resource

MusicDin:
Fixes https://github.com/terraform-lxd/terraform-provider-lxd/issues/487

### DIFF
--- a/docs/resources/trust_token.md
+++ b/docs/resources/trust_token.md
@@ -1,0 +1,60 @@
+# lxd_trust_token
+
+The `lxd_trust_token` resource allows you to request a new trust token.
+
+~> **Note:** The LXD trust token resource cannot be used for the initial authentication
+  with the LXD server because LXD Terraform provider needs to be authenticated in order
+  to request trust tokens for other clients.
+
+## Example Usage
+
+```hcl
+resource "lxd_trust_token" "token1" {
+  name = "mytoken"
+}
+
+output "token" {
+  value = lxd_trust_token.token1.token
+}
+```
+
+## Argument Reference
+
+* `name` - **Required** - Name of the token.
+
+* `projects` - *Optional* - List of projects to restrict the token to.
+
+* `remote` - *Optional* - The remote in which the resource will be created. If not provided,
+  the provider's default remote will be used.
+
+* `trigger` - *Optional* - When to trigger the token generation. Possible values are `once` and `always` (if missing). Defaults to `once`.
+
+## Attribute Reference
+
+The following attributes are exported:
+
+* `token` - The generated token.
+
+* `expires_at` - Time at which the trust token expires. If token expiry is configured, the value will be in format `YYYY/MM/DD hh:mm TZ`.
+
+## Trust token expiry
+
+~> **Warning:** The provider is unable to differentiate between an expired and a consumed token.
+  If token generation is set to `once` and the token expires, it will not be regenerated.
+
+Trust token expiry is defined in the server's configuration (`core.remote_token_expiry`).
+If the setting is configured, `expires_at` attribute will be populated, otherwise, it will be empty.
+
+For example, to set the token expiry to 20 minutes, run the following command:
+```
+lxc config set core.remote_token_expiry=20M
+```
+
+If trigger is set to `once` the token will not be regenerated on subsequent plan applies.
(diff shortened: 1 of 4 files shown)


MusicDin:
Added *trigger* for the token, so that it is generated only once by default. Other value (currently) is `always` which will ensure the token is present on apply - generated if missing.

I don't like the word *trigger*, so ideas for better naming would be appreciated (maybe `generate` or `require`?)

docs/resources/trust_token.md
```diff
@@ -0,0 +1,57 @@
...
+## Argument Reference
+
+* `name` - **Required** - Name of the token.
+
+* `projects` - *Optional* - List of projects to restrict the token to.
+
+* `remote` - *Optional* - The remote in which the resource will be created. If not provided,
+  the provider's default remote will be used.
+
+* `trigger` - *Optional* - When to trigger the token generation. Possible values are `once` and `always` (if missing). Defaults to `always`.
```
simondeziel, line 26: Please have the default to `once`.

docs/resources/trust_token.md
```diff
@@ -0,0 +1,57 @@
+# lxd_trust_token
+
+The `lxd_trust_token` resource allows you to create new tokens in the LXD trust store.
```
simondeziel, line 3: I'm not sure the tokens live in the trust store. They are a way to get your cert into the trust store though.

docs/resources/trust_token.md
```diff
@@ -0,0 +1,57 @@
+# lxd_trust_token
```
simondeziel, line 1: If you could clearly convey that those `lxd_trust_token` are meant to be issued by the LXD TF provider for the benefit of other clients wanting to interact with LXD rather than be used by the LXD TF provider itself.

docs/resources/trust_token.md
```diff
@@ -0,0 +1,57 @@
...
+The following attributes are exported:
+
+* `token` - The generated token.
+
+* `expires_at` - Time when trust token expires. If token expiry is configured, the value will be in format `YYYY/MM/DD hh:mm TZ`.
+
+## Trust token expiry
+
+~> **Warning**
+  The provider is unable to differentiate between the expired and consumed token.
```
simondeziel, line 39:
```suggestion
  The provider is unable to differentiate between an expired and a consumed token.
```

docs/resources/trust_token.md
```diff
@@ -0,0 +1,57 @@
...
+
+* `trigger` - *Optional* - When to trigger the token generation. Possible values are `once` and `always` (if missing). Defaults to `always`.
+
+## Attribute Reference
+
+The following attributes are exported:
+
+* `token` - The generated token.
+
+* `expires_at` - Time when trust token expires. If token expiry is configured, the value will be in format `YYYY/MM/DD hh:mm TZ`.
```
simondeziel, line 34:
```suggestion
* `expires_at` - Time at which the trust token expires. If token expiry is configured, the value will be in format `YYYY/MM/DD hh:mm TZ`.
```

docs/resources/trust_token.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Trust token expiry
+
+~> **Warning**
+  The provider is unable to differentiate between the expired and consumed token.
+  If token generation is set to `once` and the token expires, it will not be regenerated.
+
+Trust token expiry is defined by the system configuration `core.remote_token_expiry`.
```
simondeziel, line 42:
```suggestion
Trust token expiry is defined server-side (`core.remote_token_expiry`).
```

docs/resources/trust_token.md
```diff
@@ -0,0 +1,57 @@
...
+If the setting is configured, `expires_at` attribute will be populated, otherwise, it will be empty.
+
+For example, to set the token expiry to 20 minutes, run the following command:
+```
+lxc config set core.remote_token_expiry=20M
+```
+
+If trigger is set to `once` the token will not be regenerated on consequite applies.
```
simondeziel, line 50:
```suggestion
If trigger is set to `once` the token will not be regenerated on subsequent plan applies.
```

MusicDin:
Thanks for the comments 🚀

simondeziel (approved):
LGMT, thanks!

PR #518, opened 2024-08-23
Refactor provider config and test trust token/password

MusicDin:
The main goal is to make the provider config logic more intuitive and straight forward. This includes removal of the remote *bootstraping* during initial connection with the remote.

The majority of provider's initialization logic is also moved into `provider-config`, which makes the provider block easier to test. Provider tests are also included in this PR.

Finally, parsing of LXD remote address has been changed. Remote configuration will from now on accept only `address` attribute. The address value is expected to be a fully qualified LXD or Simplestreams address. The remote attributes `port` and `scheme` (and respective environment variables LXD_SCHEME and LXD_PORT) are now deprecated and provider will error out if set.

### DIFF
--- a/internal/acctest/checks.go
+++ b/internal/acctest/checks.go
@@ -2,10 +2,13 @@ package acctest
 
 import (
 	"fmt"
+	"net"
 	"os/exec"
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/canonical/lxd/shared/api"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"github.com/terraform-lxd/terraform-provider-lxd/internal/utils"
@@ -37,7 +40,7 @@ func PreCheckLxdVersion(t *testing.T, versionConstraint string) {
 	serverVersion := apiServer.Environment.ServerVersion
 	ok, err := utils.CheckVersion(serverVersion, versionConstraint)
 	if err != nil {
-		t.Fatal(err)
+		t.Fatalf("Failed to check LXD server version: %v", err)
 	}
 
 	if !ok {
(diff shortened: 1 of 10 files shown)


internal/acctest/testutils.go
```diff
@@ -38,3 +39,14 @@ func QuoteStrings(slice []string) string {
 	}
 	return strings.Join(quoted, ", ")
 }
+
+// ResetLXDEnvVars unsets all environment variables that are supported by
+// the provider.
+func ResetLXDEnvVars() {
+	os.Unsetenv("LXD_REMOTE")
+	os.Unsetenv("LXD_SCHEME")
```
simondeziel, line 47: Let's drop `LXD_SCHEME` and `LXD_PORT`

internal/acctest/checks.go
```diff
@@ -110,6 +111,77 @@ func PreCheckRoot(t *testing.T) {
...
+func PreCheckLocalServerHTTPS(t *testing.T) {
+	nc, err := exec.LookPath("nc")
+	if err != nil {
+		t.Skip(`Skipping remote provider test. Command "netcat" not found`)
+	}
+
+	err = exec.Command(nc, "-z", "127.0.0.1", "8443").Run()
+	if err != nil {
+		t.Skip(`Skipping remote provider test. LXD is not available on "https://127.0.0.1:8443"`)
+	}
```
simondeziel, line 125: Please do that TCP open check in go.

internal/acctest/testutils.go
MusicDin: Done :)

internal/acctest/checks.go
MusicDin: Yeah really good point. Thanks, don't know what I was thinking.. 👯

internal/provider-config/config.go
```diff
@@ -266,185 +350,213 @@ func (p *LxdProviderConfig) createLxdServerClient(remote LxdProviderRemoteConfig
...
+func (p *LxdProviderConfig) setRemote(remoteName string, remote LxdRemote) error {
+	p.mux.Lock()
+	defer p.mux.Unlock()
 
-	certificate, err := lxd_shared.GetRemoteCertificate(lxdRemote.Addr, "terraform-provider-lxd/2.0")
-	if err != nil {
-		return err
+	// Validate remote.
+	if remoteName == "" {
+		return fmt.Errorf("Remote name cannot be empty")
```
simondeziel, line 393:
```suggestion
		return errors.New("Remote name cannot be empty")
```
Should we use what we do in LXD when there is no format specifier?

internal/provider-config/config.go
```diff
@@ -3,76 +3,208 @@ package config
 import (
 	"encoding/pem"
```
simondeziel, line 4: Missing `"errors"`

simondeziel (approved):
Thank you!

PR #509, opened 2024-08-07
Don't check fingerprint if 'Content' is unknown

jsimpso:
Closes #507

PR #519, opened 2024-08-26
Test truststore certificate using cert generated within the same TF config

MusicDin:
Adds test for #509


simondeziel:
> Adds test for #509

Funny how that PR number is related with X.509 certificates ;)

simondeziel (approved):

PR #526, opened 2024-09-20
Allow instance rename

MusicDin:
Instead of replacing an instance, allow instance to be renamed. This is only possible if instance is either currently stopped or the `running` field is set to `false`.

### DIFF
--- a/internal/instance/resource_instance_test.go
+++ b/internal/instance/resource_instance_test.go
@@ -216,6 +216,62 @@ func TestAccInstance_restartVirtualMachine(t *testing.T) {
 	})
 }
 
+func TestAccInstance_renameInstance(t *testing.T) {
+	instanceNameA := acctest.GenerateName(3, "-")
+	instanceNameB := acctest.GenerateName(3, "-")
+	instanceNameC := acctest.GenerateName(3, "-")
+	instanceNameD := acctest.GenerateName(3, "-")
+	instanceNameE := acctest.GenerateName(3, "-")
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctest.PreCheck(t) },
+		ProtoV6ProviderFactories: acctest.ProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				// Launch a new instance.
+				Config: testAccInstance_rename(instanceNameA, true),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("lxd_instance.instance1", "name", instanceNameA),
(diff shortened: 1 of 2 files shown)


internal/instance/resource_instance_test.go
```diff
@@ -216,6 +216,55 @@ func TestAccInstance_restartVirtualMachine(t *testing.T) {
...
+				Config: testAccInstance_rename(instanceNameB, true),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("lxd_instance.instance1", "name", instanceNameB),
+					resource.TestCheckResourceAttr("lxd_instance.instance1", "status", "Running"),
+				),
+			},
+			{
+				// Ensure instance rename fails when instance is running.
+				Config:      testAccInstance_rename(instanceNameA, true),
+				ExpectError: regexp.MustCompile("Renaming of running instance not allowed"),
```
simondeziel, line 262: Should we also have the `resource.TestCheckResourceAttr("lxd_instance.instance1", "name", instanceNameB),` checked here?

internal/instance/resource_instance_test.go
```diff
@@ -216,6 +216,55 @@ func TestAccInstance_restartVirtualMachine(t *testing.T) {
...
+				// Launch a new instance.
+				Config: testAccInstance_rename(instanceNameA, true),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("lxd_instance.instance1", "name", instanceNameA),
+					resource.TestCheckResourceAttr("lxd_instance.instance1", "status", "Running"),
+				),
+			},
+			{
+				// Stop and rename the instance.
+				Config: testAccInstance_rename(instanceNameB, false),
```
simondeziel, line 237: The provided name could be a bit more indicative of the intention. Say the instance started with `A`, then move to `A2B` -> `B2C` -> `C2D` -(fail)-> `D2E`.

Just an idea as I think what you have is fine, it just required a bit more mental tracking.

simondeziel (approved):

internal/instance/resource_instance_test.go
MusicDin: Not sure if state is testable when error is expected, but worth trying.

internal/instance/resource_instance_test.go
MusicDin: Hehe, will fix :)

internal/instance/resource_instance_test.go
MusicDin: Seems to be working. Thanks :)

simondeziel (approved):

Issue #405, opened 2023-12-20
Please import storage bucket support from provider incus


simondeziel:
https://github.com/lxc/terraform-provider-incus/pull/5

PR #527, opened 2024-09-20
Add storage bucket resource

MusicDin:
Closes https://github.com/terraform-lxd/terraform-provider-lxd/issues/405

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,6 +63,23 @@ jobs:
           # Add HTTPS remote.
           lxc remote add localhost "$(lxc config trust add --name lxd-terraform-provider --quiet)"
 
+      - name: Configure MinIO
+        run: |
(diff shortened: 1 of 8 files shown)


.github/workflows/test.yml
```diff
@@ -63,6 +63,23 @@ jobs:
           # Add HTTPS remote.
           lxc remote add localhost "$(lxc config trust add --name lxd-terraform-provider --quiet)"
 
+      - name: Configure MinIO
+        run: |
+          arch=$(dpkg --print-architecture)
+          mkdir -p /opt/minio
```
simondeziel, line 69: It's surprising (to me) that `/opt` is writable by the runner user. Maybe a special thing on GHA runners... This is normally root writable only.

Doesn't matter, just surprising :)

simondeziel (approved):

Issue #520, opened 2024-08-30
TF should not allow accepting the LXD server's cert when using a token


simondeziel:
Similar to https://github.com/canonical/lxd/issues/13973, when a token is used, TF should not blindly accept the server's cert it is presented. The fingerprint contained in the token should always be used to ensure we get the right cert.

PR #529, opened 2024-09-20
Do not blindly accept remote certificate when using trust token

MusicDin:
Closes https://github.com/terraform-lxd/terraform-provider-lxd/issues/520

### DIFF
--- a/internal/provider-config/config.go
+++ b/internal/provider-config/config.go
@@ -320,7 +320,7 @@ func (p *LxdProviderConfig) server(remoteName string) (lxd.Server, error) {
 			}
 
 			// Try to accept the remote certificate.
-			err := p.acceptRemoteCertificate(remoteName, remote.Address)
+			err := p.acceptRemoteCertificate(remoteName, remote.Token, remote.Address)
 			if err != nil {
 				return nil, fmt.Errorf("Failed to accept server certificate for remote %q: %v", remoteName, err)
 			}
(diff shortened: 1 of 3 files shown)


simondeziel (approved):

simondeziel:
https://github.com/terraform-lxd/terraform-provider-lxd/actions/runs/10962720240/job/30442627935?pr=529#step:8:339:
> certificate signed by unknown authority

Sounds like its trying to do "regular" TLS verification without having the remote's `.crt` pinned/saved in the local trust store and used as trust anchor.

internal/provider-config/config.go
```diff
@@ -442,6 +444,20 @@ func (p *LxdProviderConfig) acceptRemoteCertificate(remoteName string, url strin
...
+	if token != "" {
+		// Decode token.
+		rawToken, err := shared.CertificateTokenDecode(token)
+		if err != nil {
+			return fmt.Errorf("Failed decoding trust token for remote %q: %v", remoteName, err)
+		}
+
+		// Compare token and certificate fingerprints.
+		certFingerprint := shared.CertFingerprint(cert)
+		if certFingerprint != rawToken.Fingerprint {
```
simondeziel, line 456: That feels like something the LXD client should be doing itself, no?

Doing the token decoding + verification in the provider feels like a layer violation to me. I guess my intention was to simply error out if a token was provided along with asking to blindly accept the remote cert. In pseudo-code that'd be:

```
if token != "" && p.acceptServerCertificate {
    return errors.New("Using a remote add token is incompatible with accept_remote_certificate=true".)
}
```

MusicDin: Hmm, but then the user would need to have a server certificate already present locally?

The current approach makes the `acceptRemoveCertificate` flag redundant when using tokens, as the remote certificate fingerprint will be always compared to the one within the token. Therefore, whenever the fingerprint matches, the certificate is accepted.

internal/provider-config/config.go
```diff
@@ -428,9 +428,11 @@ func (p *LxdProviderConfig) setRemote(remoteName string, remote LxdRemote) error
...
-	// Check if we are allowed to accept the remote certificate.
-	if !p.acceptServerCertificate {
+func (p *LxdProviderConfig) acceptRemoteCertificate(remoteName string, token string, url string) error {
+	// Check if we are allowed to blindly accept the remote certificate.
+	// When the trust token is used, the fingerprint contained in the token
+	// is used to ensure we get the right certificate.
+	if token == "" && !p.acceptServerCertificate {
 		return errors.New("Unable to communicate with remote server. " +
 			`Either set "accept_remote_certificate" to true or add ` +
 			"the remote out of band of Terraform and try again.")
```
simondeziel, line 438: Here again we should recommend using a remote add token as well.

internal/provider-config/config.go
```diff
@@ -442,6 +444,20 @@ func (p *LxdProviderConfig) acceptRemoteCertificate(remoteName string, url strin
 		return err
 	}
 
+	if token != "" {
+		// Decode token.
+		rawToken, err := shared.CertificateTokenDecode(token)
```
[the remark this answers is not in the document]
simondeziel: `s/rawToken/decodedToken/` maybe?

internal/provider-config/config.go
simondeziel: After discussing this at length, you are right, peaking at the token and doing the fingerprint comparison needs to be done by the provider because the LXD Go API doesn't provide a more straightforward way of doing this.

internal/provider-config/config.go
MusicDin: Yep, please check if it is ok now.

simondeziel (approved):

PR #545, opened 2024-11-29
provider-config: Append port to hostname instead of the URL

### LINKED ISSUE #539: Fails if remote address contains a subdirectory
# summary
The following settings will fail.
```tf
 remote {
    name    = "ubuntu-ex"
    protocol = "simplestreams"
    address = "https://example.com/ubuntu-cloud-images/releases/"
  }
```
※Domain section is hidden.

```
resource "lxd_cached_image" "image" {
    source_remote   = "ubuntu-ex"
    source_image    = "24.04"
    type            = "container"
}
```

Errors are as follows
```
│ Failed getting image: Failed parsing stream: Unable to fetch
│ https://example.com/ubuntu-cloud-images/releases/:443/streams/v1/index.json: 404 Not Found
```

https://example.com/ubuntu-cloud-images/releases/streams/v1/index.json
has been confirmed to exist.

# What you need to fix
I want the port number to come to the proper location when making http requests.

https://example.com/ubuntu-cloud-images/releases/:443/streams/v1/index.json
↓
https://example.com:443/ubuntu-cloud-images/releases/streams/v1/index.json

MusicDin:
Closes #539

### DIFF
--- a/internal/provider-config/config.go
+++ b/internal/provider-config/config.go
@@ -569,10 +569,10 @@ func DetermineLXDAddress(protocol string, address string) (string, error) {
 				port = "443"
 			}
 
-			address = fmt.Sprintf("%s:%s", address, port)
+			url.Host = fmt.Sprintf("%s:%s", url.Hostname(), port)
 		}
 
-		return address, nil
+		return url.String(), nil
 	default:
 		return "", fmt.Errorf("Invalid scheme %q: Value must be one of: [unix, https]", scheme)
 	}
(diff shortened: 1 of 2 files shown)


internal/provider-config/config.go
```diff
@@ -569,10 +569,10 @@ func DetermineLXDAddress(protocol string, address string) (string, error) {
 				port = "443"
 			}
 
-			address = fmt.Sprintf("%s:%s", address, port)
+			url.Host = fmt.Sprintf("%s:%s", url.Hostname(), port)
```
simondeziel, line 572: Since we are dealing with strings I'd avoid the call to `Sprintf`:
```suggestion
			url.Host = url.Hostname() + ":" + port
```
But it's not performance critical ;)

simondeziel (approved):

PR #548, opened 2024-12-10
Gomod updates

simondeziel:


Makefile
```diff
@@ -65,7 +65,7 @@ static-analysis:
 .PHONY: update-gomod
 update-gomod:
 	$(GO) get -t -v -u ./...
-	$(GO) mod tidy --go=1.22.3
+	$(GO) mod tidy --go=1.23.3
```
MusicDin, line 68: Is that flag to enforce specific go version? Do we need that flag?

Issue #523, opened 2024-08-30
Add Golang linter in CI


MusicDin:
It would be nice if we add go linter (similar to one in LXD). For example, to detect redundant "else" statements, unused parameters, etc.

simondeziel:
@MusicDin re-assigning to you so that you can slowly renable the currently disabled `revive` rules.

PR #549, opened 2024-12-10
Improve golangci-lint coverage

simondeziel:
Partly addresses #523.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,23 +18,23 @@ linters-settings:
       - (*os.File).Close
   revive:
     rules:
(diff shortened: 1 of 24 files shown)


simondeziel:
@MusicDin I fixed the low hanging fruit (with your help, thanks!) but ended up disabling the `exported` and `var-naming` rules from `revive`. Would you be willing to make them work? I feel like knowing the code base much better than I do means you are in a better position to do this.

MusicDin (commented):
Thanks! 🚀

MusicDin (approved):

PR #553, opened 2024-12-20
github: Fix MinIO install step condition

MusicDin:
Fix MinIO install step condition and some other minor cosmetic fixes in GH workflows.


simondeziel (approved):
Thanks!

Issue #546, opened 2024-11-29
Error when “@cluster_group” is specified for target


tam1192:
# Error Description
```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to lxd_instance.backup, provider "provider[\"registry.terraform.io/terraform-lxd/lxd\"]" produced an unexpected new value: .target: was cty.StringVal("@gpu"), but now cty.StringVal("node-2").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

## environment
```
terraform version
Terraform v1.6.3-dev
on linux_amd64
+ provider registry.terraform.io/terraform-lxd/lxd v2.4.0

Your version of Terraform is out of date! The latest version
is 1.9.8. You can update by downloading from https://www.terraform.io/downloads.html
```

```
uname -a
Linux nikki-main 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 GNU/Linux
```

# what should happen?
No more errors.

# What we tried
Tentatively specifying by node name avoids errors.
If there's anything you'd like me to try, let me know.

MusicDin:
Hmm, yeah that one is interesting. 

Could you check the result after failure? I'm guessing the instance was created within the expected group (`@gpu`), but the provider updates the state with the actual instance location which causes the inconsistent result.

If this is the case, we just need to introduce the field `location` which would track exact location, and `target` would be used just for user input (as we already do on some resources).

tam1192:
Sorry for the delay in replying.
I get the above error and the terraform process ends, though,
In fact, an instance exists.

Here is its condition.

```
lxc ls
+---------------+---------+-----------------------+-------------------------------------------------+-----------------+-----------+----------+
|     NAME      |  STATE  |         IPV4          |                      IPV6                       |      TYPE       | SNAPSHOTS | LOCATION |
+---------------+---------+-----------------------+-------------------------------------------------+-----------------+-----------+----------+
| ex-instance-0 | RUNNING | 192.168.1.10 (enp5s0) | fd42:f573:2540:b54d:216:3eff:fe2d:ee46 (enp5s0) | VIRTUAL-MACHINE | 0         | node-2   |
+---------------+---------+-----------------------+-------------------------------------------------+-----------------+-----------+----------+
```

PR #559, opened 2025-02-17
Allow cluster member group as target and handle in-cluster migration

MusicDin:
Previously, the provider updated `target` attribute with the instance's actual location, which caused plan mismatch if cluster member group was set as target. This PR changes the instance `target` attribute to only accept user value and reports the current instance location in `location` attribute instead.

Additionally, when instance is misplaced, provider now attempts to migrate the instance to the desired location (within a cluster) instead of replacing it. It also introduces a new attribute `allow_restart` which defaults to false. If enabled, provider will stop the instance before attempting migration or instance rename, and start it again once finished. If the instance is already stopped or planned to be stopped, the provider will handle the operation correctly regardless of `allow_restart` value.

Fixes https://github.com/terraform-lxd/terraform-provider-lxd/issues/546

### DIFF
--- a/internal/instance/resource_instance.go
+++ b/internal/instance/resource_instance.go
@@ -42,6 +42,7 @@ type InstanceModel struct {
 	Image          types.String `tfsdk:"image"`
 	Ephemeral      types.Bool   `tfsdk:"ephemeral"`
 	Running        types.Bool   `tfsdk:"running"`
+	AllowRestart   types.Bool   `tfsdk:"allow_restart"`
 	WaitForNetwork types.Bool   `tfsdk:"wait_for_network"`
 	Profiles       types.List   `tfsdk:"profiles"`
 	Devices        types.Set    `tfsdk:"device"`
@@ -126,6 +127,13 @@ func (r InstanceResource) Schema(ctx context.Context, _ resource.SchemaRequest, 
 				Default:  booldefault.StaticBool(true),
 			},
 
+			"allow_restart": schema.BoolAttribute{
+				Description: "Allow instance to be stopped and restarted if required by the provider for operations like migration or renaming.",
+				Optional:    true,
+				Computed:    true,
+				Default:     booldefault.StaticBool(false),
+			},
+
 			"wait_for_network": schema.BoolAttribute{
 				Optional: true,
 				Computed: true,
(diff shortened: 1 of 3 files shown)


internal/instance/resource_instance.go
```diff
@@ -1327,10 +1394,65 @@ func stopInstance(ctx context.Context, server lxd.InstanceServer, instanceName s
...
+	// is empty, we do not really care where the instance is located.
+	if !server.IsClustered() || target == "" {
+		return true, nil
+	}
+
+	// If target has prefix "@", we are deling with the cluster
+	// member group.
+	targetGroup, ok := strings.CutPrefix(target, "@")
+	if ok {
+		members, err := server.GetClusterMembers()
```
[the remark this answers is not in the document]
MusicDin: Yes, we can just fetch the group directly and check if the current location equals to one of the group members. Thanks.

internal/instance/resource_instance_test.go
```diff
@@ -1267,6 +1276,103 @@ func TestAccInstance_timeout(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("lxd_instance.instance1", "name", instanceName),
+					resource.TestCheckResourceAttr("lxd_instance.instance1", "status", "Running"),
+					resource.TestCheckResourceAttr("lxd_instance.instance1", "target", targets[0]),
+					resource.TestCheckResourceAttr("lxd_instance.instance1", "location", targets[0]),
+				),
+			},
+			{
+				// Set instance target to cluster group "@default" (noop).
+				// All cluster members are part of default group.
```
simondeziel, line 1303: It would be nice to use a group that causes the relocation of the instance.

internal/instance/resource_instance.go
```diff
@@ -1426,3 +1512,40 @@ func findGlobalIPAddresses(network api.InstanceStateNetwork) (ipv4 string, ipv6
...
+		// If the current cluster member (location) is part of the target
+		// cluster group, then the instance is located on the correct
+		// cluster member.
+		if slices.Contains(group.Members, location) {
+			return true, nil
+		}
+	}
+
+	// The instance is not located on the desired cluster member
+	// or within a desired cluster member group.
```
simondeziel, line 1549:
```suggestion
	// The instance is not located on the desired cluster member/group.
```

simondeziel (approved):

simondeziel (approved):

PR #563, opened 2025-02-23
feat: allow creation of empty VMs

### LINKED ISSUE #561: Feature Request: Allow a LXD instance to use an 'empty' image
The use of `--empty` is handy when you want to test booting from an alternative source such as PXE booting, or booting from an ISO. Both of which I do frequently. Most of my lab is automated via Terraform (great work on the provider!), but these steps still need to be done manually.  Is it possible to implement this feature so I can do something like the following?

```yaml
name: medium
description: 2 CPUs, 4GiB RAM, 15GiB DISK
config:
  limits.cpu: "2"
  limits.memory: 4GiB
devices:
  eth0:
    boot.priority: "5"
    network: pxe-net
    type: nic
  root:
    boot.priority: "10"
    path: /
    pool: default
    size: 15GiB
    type: disk
used_by:
- /1.0/instances/test?project=pxe-repros
```
```bash
lxc launch --vm --empty -p medium pxe-boot
```

Likewise, the Terraform might look something like this:
```hcl
resource "lxd_profile" "flavor" {
  for_each = local.flavors

  name = each.key
  description = each.value.description
  project = lxd_project.pxe.name

  config = {
    "limits.cpu" = each.value.cpu
    "limits.memory" = each.value.memory
  }

  # Boot priority is set lower than root disk
  # It will PXE boot first time with an empty disk, then boot from the root
  device {
    name = "eth0"
    type = "nic"

    properties = {
      "boot.priority" = "5"
      network = lxd_network.pxe_net.name
    }
  }

  device {
    name = "root"
    type = "disk"

    properties = {
      "boot.priority" = "10"
      pool = var.storage_name
      path = "/"
      size = each.value.size
    }
  }
}

resource "lxd_instance" "pxe" {
  name = "boot-me"
  project = lxd_project.pxe.name
  profiles = [lxd_profile.pxe_flavor["medium"].name]
  # Can set image = "empty", "", null, or simply not make this parameter required. However, if this is empty, it breaks containers. Validation should be done for it only applies to virtual-machines
  image = "empty"
  type = "virtual-machine"
  running = false

  # Deploy if it is supposed to
  count = var.pxe_deploy ? 1 : 0

  timeouts = {
    create = "30m"
  }
}
```

Please keep in mind empty images will break containers,  but are perfectly valid for virtual-machines. Validation should be done if possible to ensure that's only the case with virtual-machines

xvzf:
Note: the `TestAccInstance_empty_vm` is skipped due to the LXD agent not
coming up (since there's no OS booted). Open for ideas on how to test
this efficiently. My only ideas have been PXE booting in a pre-existing
environment or attaching a custom ISO volume. Both of them are not
really portable.

fixes #561

Signed-off-by: Matthias Riegler <github@m4tbit.de>

### DIFF
--- a/docs/resources/instance.md
+++ b/docs/resources/instance.md
@@ -47,8 +47,8 @@ resource "lxd_instance" "instance2" {
 
 * `name` - **Required** - Name of the instance.
 
-* `image` - **Required** - Base image from which the instance will be created. Must
-  specify [an image accessible from the provider remote](https://documentation.ubuntu.com/lxd/en/latest/reference/remote_image_servers/).
+* `image` - **Optional** - Base image from which the instance will be created. **For containers** it must
+  specify [an image accessible from the provider remote](https://documentation.ubuntu.com/lxd/en/latest/reference/remote_image_servers/). If omitted, it is equivalent to the `--empty` CLI flag and creates an empty virtual machine.
 
 * `description` - *Optional* - Description of the instance.
 
--- a/internal/instance/resource_instance.go
+++ b/internal/instance/resource_instance.go
@@ -106,7 +106,7 @@ func (r InstanceResource) Schema(ctx context.Context, _ resource.SchemaRequest, 
 			},
 
 			"image": schema.StringAttribute{
-				Required: true,
+				Optional: true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplace(),
 				},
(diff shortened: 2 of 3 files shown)


xvzf:
gonna fix the linting

stew3254:
Thanks @xvzf! I'm looking forward to using this. Looks like it wasn't too hard.

MusicDin:
Thanks, how did you test it manually? I would like to include some basic tests to prevent potential regression in the future.

simondeziel (changes requested):
LGTM

xvzf:
> Thanks, how did you test it manually? I would like to include some basic tests to prevent potential regression in the future.

I created a virtual machine which PXE booted (in my case) -- but I'm also working on another PR for this project to import ISOs from an http/file source

MusicDin:
Could you please sign your last commit?

xvzf:
@MusicDin should be done now!

MusicDin (approved):
Thanks

Issue #569, opened 2025-03-12
Taking an instance snapshot with an invalid name results in a segfault


stew3254:
It appears you cannot take volume snapshots, only instance snapshots. Just to test my theory, I decided to try to snapshot a volume I created and ended up with a segfault. I expected the volume snapshot to fail because that instance name doesn't exist, but I was not expecting a segfault. Now it's only partly executed my Terraform plan before it crashed.

```
module.student-lab["student-lab"].lxd_snapshot.kali_home["stew3254"]: Creating...
╷
│ Error: Request cancelled
│ 
│ The plugin6.(*GRPCProvider).ApplyResourceChange request was cancelled.
╵
╷
│ Error: Request cancelled
│ 
│ The plugin6.(*GRPCProvider).ApplyResourceChange request was cancelled.
╵

Stack trace from the terraform-provider-lxd_v2.4.0 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xc6242c]

goroutine 109 [running]:
github.com/terraform-lxd/terraform-provider-lxd/internal/instance.InstanceSnapshotResource.Create({0x0?}, {0x1056918, 0xc000817e90}, {{{{0x105cc90, 0xc00034d260}, {0xd8f360, 0xc00034d0b0}}, {0x105f418, 0xc0002ae9b0}}, {{{0x105cc90, ...}, ...}, ...}, ...}, ...)
	github.com/terraform-lxd/terraform-provider-lxd/internal/instance/resource_instance_snapshot.go:149 +0x6ec
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).CreateResource(0xc0002481e0, {0x1056918, 0xc000817e90}, 0xc000395538, 0xc000395510)
	github.com/hashicorp/terraform-plugin-framework@v1.12.0/internal/fwserver/server_createresource.go:101 +0x578
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).ApplyResourceChange(0xc0002481e0, {0x1056918, 0xc000817e90}, 0xc0000bebe0, 0xc000395628)
	github.com/hashicorp/terraform-plugin-framework@v1.12.0/internal/fwserver/server_applyresourcechange.go:57 +0x4aa
[...]

PR #572, opened 2025-03-12
Fix segfault when taking instance snapshot for an instance doesn't exist

stew3254:
This PR addresses https://github.com/terraform-lxd/terraform-provider-lxd/issues/569 which fixes the segfault when taking instance snapshot for an instance doesn't exist. It also implements a unit test so that we can see expected behavior

### DIFF
--- a/internal/instance/resource_instance_snapshot_test.go
+++ b/internal/instance/resource_instance_snapshot_test.go
@@ -2,6 +2,7 @@ package instance_test
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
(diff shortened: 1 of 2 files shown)


internal/instance/resource_instance_snapshot_test.go
```diff
@@ -191,3 +208,12 @@ resource "lxd_snapshot" "snapshot1" {
 }
 	`, project, instance, acctest.TestImage, snapshot)
 }
+
+func testAccInstanceSnapshot_missingInstance(cName, sName string) string {
```
MusicDin, line 212: Just a small nit, but could you please update the `cName, sName string` -> `instanceName string, snapshotName string` on the test config functions? Seems those have slipped through.

MusicDin (commented):
Thanks, looks good. One optional *cosmetic* comment.

Could you just sign your commits please?

internal/instance/resource_instance_snapshot_test.go
stew3254: Sure

MusicDin (approved):
Thanks!

PR #584, opened 2025-04-28
Add instance device resource

nmezhenskyi:
This PR adds `InstanceDeviceResource` that allows attaching devices to an instance after it was created. This can be particularly useful for attaching volumes to an instance with initially unknown location in the cluster.

**TODO**:

- [x] Add `user.managed-by: terraform-lxd-provider` config values to devices managed by Terraform to differentiate between infrastructure vs manually added devices.
- [x] Remove any manually added devices.
- [ ] Add tests for coexisting devices in the `instance` resource definition and in `device` resource.
- [ ] Add tests for clustered volume attachment to instances.

### DIFF
--- a/docs/resources/instance.md
+++ b/docs/resources/instance.md
@@ -364,3 +364,7 @@ import {
   state and treated as computed values.
     - `image.*`
     - `volatile.*`
+
+* Terraform LXD provider sets user.managed-by key to all managed instance devices.
+  Removing that key from a device manually, would result in Terraform removing it on next apply.
+
--- a/docs/resources/instance_device.md
+++ b/docs/resources/instance_device.md
@@ -1,0 +1,87 @@
+# lxd_instance_device
+
+Manages a single device in an LXD instance. Provides an alternative way to attach
+and manage a device.
+
+This resource is useful for managing devices on a LXD instance in a cluster.
+For example, it allows attaching volumes to an instance with initially unknown
+location in the cluster.
+
+~> **Warning:** This is an experimental feature of Terraform LXD Provider and it may
+   change in the future.
+
+## Example
+
+```hcl
+resource "lxd_instance" "instance1" {
+  name  = "instance1"
+  image = "ubuntu-daily:22.04"
+
+  # Device attached during instance creation.
+  device {
+    name = "eth0"
+    type = "nic"
+
+    properties = {
+      network = "lxdbr0"
+    }
+  }
+}
+
+resource "lxd_volume" "vol1" {
+  name   = "vol1"
+  pool   = "default"
+  type   = "disk"
+
+  # In case we want to create the volume on the same cluster member
+  # where the instance has been created, the "target" needs to match.
+  #
+  # However, this prevents us from using device block in instance
+  # resource, as it would create a dependency loop. Therefore, we
+  # will use "lxd_instance" resource, which allows device attachment
+  # after instance creation.
+  target = lxd_instance.instance1.location
+
+  config {
+    size = "10GB"
+  }
+}
+
+# Device attachment after instance creation.
+resource "lxd_device" "vol1" {
+  instance     = lxd_instance.instance1.name # Target instance.
+  device_name  = lxd_volume.vol1.name        # Target volume, which is created after instance creation.
+  type         = "disk"
+
+  properties = {
+    path   = "/mnt/vol1"
+    pool   = "default"
+    source = lxd_volume.vol1.name
+  }
+}
+```
+
+## Argument Reference
+
+* `name` - **Required** - Name of the device.
+
+* `instance_name` - **Required** - Name of the instance.
+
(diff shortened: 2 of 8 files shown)


internal/device/resource_device.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+
+	m.Type = types.StringValue(deviceType)
+	m.Properties = properties
+
+	return tfState.Set(ctx, &m)
+}
+
+// createInstancePut creates api.InstancePut struct required for instance update.
+func createInstancePut(instance *api.Instance) api.InstancePut {
```
MusicDin, line 382: You should be able to use `instance.Writable()` which returns the `api.InstancePut` structure

internal/device/resource_device.go
```diff
@@ -0,0 +1,392 @@
...
+	deviceName := plan.Name.ValueString()
+	deviceType := plan.Type.ValueString()
+	props["type"] = deviceType
+
+	if instance.Devices == nil {
+		instance.Devices = make(map[string]map[string]string)
+	}
+
+	// Modify devices map to add the provided device.
+	instance.Devices[deviceName] = props
```
MusicDin, line 186: What happens if the device already exists, and a new device is configured through the provider? I think we should error out in `Create()` function if the device already exists to prevent unintentional modification.

internal/device/resource_device.go
```diff
@@ -0,0 +1,392 @@
...
+
+	// Modify devices map to remove specified device.
+	delete(instance.Devices, deviceName)
+
+	updatedInstance := createInstancePut(instance)
+	op, err := server.UpdateInstance(instance.Name, updatedInstance, etag)
+	if err == nil {
+		// Wait for the instance to be updated.
+		err = op.WaitContext(ctx)
+	}
```
MusicDin, line 322: Can we skip the update if the device is not found in the retrieved instance?

internal/device/resource_device.go
```diff
@@ -0,0 +1,392 @@
...
+
+	deviceType, ok := deviceProps["type"]
+	if !ok || deviceType == "" {
+		respDiags.AddError(
+			"Device is missing type",
+			fmt.Sprintf("Device %q for instance %q is missing type field", deviceName, instanceName))
+		return respDiags
+	}
+	delete(deviceProps, "type")
```
MusicDin, line 366: Why are we deleting `type`? This could benefit from a comment

MusicDin (changes requested):
Beside some minor comments, already looking very good 🚀

internal/device/resource_device.go
nmezhenskyi: Yes, makes sense, we don't want to overwrite the device in `Create()`. I'll add a check for it.

internal/device/resource_device.go
nmezhenskyi: Yes, makes sense

internal/device/resource_device.go
nmezhenskyi: Should we also error out in `Update()` function if the device does not exist?

internal/device/resource_device.go
nmezhenskyi: I've added a comment in the code

internal/device/resource_device.go
MusicDin: Yes please

internal/device/resource_device_test.go
```diff
@@ -0,0 +1,197 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("lxd_storage_pool.pool1", "name", poolName),
+					resource.TestCheckResourceAttr("lxd_storage_pool.pool1", "driver", "zfs"),
+					resource.TestCheckResourceAttr("lxd_volume.volume1", "name", volumeName),
+					resource.TestCheckResourceAttr("lxd_volume.volume1", "pool", poolName),
+					resource.TestCheckResourceAttr("data.lxd_instance.inst", "name", instanceName),
+					resource.TestCheckResourceAttr("data.lxd_instance.inst", "status", "Stopped"),
+					resource.TestCheckResourceAttr("data.lxd_instance.inst", "devices.%", "0"),
+				),
+			},
```
nmezhenskyi, line 102: I couldn't find a better way to test that. Lifecycle post/precondition couldn't make the data source "wait" until the `resource "lxd_device" "disk_attach"` would get deleted. So, the data source would read the state when the instance still has the device attached. But splitting the validation into a separate test step just works.

docs/resources/instance.md
```diff
@@ -364,3 +364,7 @@ import {
   state and treated as computed values.
     - `image.*`
     - `volatile.*`
+
+* Terraform LXD provider sets user.managed-by key to all managed instance devices.
```
MusicDin, line 368: Format as code block -> `user.managed-by`

MusicDin (approved):
Thanks 🚀

Issue #586, opened 2025-05-07
Creation fails if lxd agent is not installed in virtual machine


tsackit:
## Problem 

I have a non-systemd vm image (Talos Linux), when I try to create the vm it always gives me an error:

```
Failed to wait for instance "TalosVM-ccd18dc" to start: timeout while waiting for state to become 'Running' (last state: 'Running (initializing)', timeout: 3m0s)
```

I have checked the code and this error is caused in the start process:

https://github.com/terraform-lxd/terraform-provider-lxd/blob/94ad83b72d0053469a66835653625f4b0897f456/internal/instance/resource_instance.go#L1360

The `Processes` property is reported by the lxd agent, but this agent is not available for non-systemd distros. My vm reports as running with `status_code = 103`.

https://github.com/terraform-lxd/terraform-provider-lxd/blob/94ad83b72d0053469a66835653625f4b0897f456/internal/instance/resource_instance.go#L1502

Can we get around this check?.

MusicDin:
The `wait_for` was introduced on instance in https://github.com/terraform-lxd/terraform-provider-lxd/pull/687, which allows to not wait for LXD agent.

PR #588, opened 2025-05-07
Add wait_for_operational to Instance resource

tsackit:
This PR allows to disable operational check, is usefull for vm that dont have `lxd-agent`

#586

TODO:
---

- [ ] Add test

### DIFF
--- a/docs/resources/instance.md
+++ b/docs/resources/instance.md
@@ -61,6 +61,9 @@ resource "lxd_instance" "instance2" {
 * `wait_for_network` - *Optional* - Boolean indicating if the provider should wait for the instance to get an IPv4 address before considering the instance as started.
   If `running` is set to false or instance is already running (on update), this value has no effect. Defaults to `true`.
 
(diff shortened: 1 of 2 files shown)


MusicDin:
Very nice, thanks!

Could you just include one container and VM test for this?

tsackit:
@MusicDin ready 🚀

MusicDin (changes requested):
Looking good, thanks.

I think we should just include two more tests (or expand the existing ones) where `wait_for_operational=false` is used with instance file upload and instance execs.

The issue that can arise here is that the instance is likely not operational when we attempt to upload files, or run commands on the instance, resulting in unexpected errors.

This could be avoided by preventing usage of `file` blocks and `execs` map when `wait_for_operational=false`. 

---

Unrelated, but if you are willing to share some examples of TalosOS and lxd-terraform-provider, I would really appreciate it. In particular, I would like to investigate it and see if we can further improve support for it. We could also include some examples in docs?

Thanks 🚀

MusicDin:
Hi @tsackit, just checking whether you are still interested in landing this change?

PR #602, opened 2025-07-07
Fix unsupported content-type during file upload

MusicDin:
Using `*bytes.Buffer`, `*bytes.Reader`, or `*strings.Reader` will make `(net/http).NewRequestWithContext` set `Content-Length` on the request to `> 0`, but the `Content-Type` will remain unset, causing LXD to detect that as inappropriate content-type during raw-content file upload.

This should unblock the tests for now.


nmezhenskyi (approved):
Looks good to me, thanks!

PR #629, opened 2025-11-04
import: Ensure colon is interpreted for remote only before first slash

### LINKED ISSUE #628: Creating IPv6 `lxd_network_forward` without explicit remote fails
### Please confirm

### Issue description

As the title states, creating IPv6 forwarders fails if the resource does not have an explicit `remote` specified with error:
```
│ Error: Missing Resource State After Create
│ 
│   with lxd_network_forward.nextcloud,
│   on forwards.tf line 1, in resource "lxd_network_forward" "nextcloud":
│    1: resource "lxd_network_forward" "nextcloud" {
│ 
│ The Terraform Provider unexpectedly returned no resource state after having no errors in the resource creation. This is always an issue in the Terraform Provider and should be reported to the provider developers.
│ 
│ The resource may have been successfully created, but Terraform is not tracking it. Applying the configuration again with no other action may result in duplicate resource errors. Import the resource if the resource was actually created and Terraform
│ should be tracking it.
```
The network forward is actually created in the LXD, the provider just fails to import its state afterwards. I believe this to be caused by the colons in the IPv6 address and the way the provider parses remote IDs
https://github.com/terraform-lxd/terraform-provider-lxd/blob/6853faa4ace05ef073e3ea72a4426fa1e9b37765/internal/common/lxd_import.go#L74-L83 

In the absence of explicitly specified `remote`, it splits the ID on the first colon of the IPv6 address instead. It can also be reproduced when manually importing the resource.

Does not work:
```
$ tofu import lxd_network_forward.nextcloud nextcloud/2001:db8::1
lxd_network_forward.nextcloud: Importing from ID "nextcloud/2001:db8::1"...
data.lxd_instance.nextcloud: Reading...
data.lxd_instance.nextcloud: Read complete after 0s [name=nextcloud]
╷
│ Error: Invalid import ID: "nextcloud/2001:db8::1"
│ 
│ Import ID does not contain all required fields: [network, listen_address].
│ 
│ Valid import format:
│ import lxd_network_forward.<resource> [<remote>:][<project>]/<network>/<listen_address>
```

Works:
```
$ tofu import lxd_network_forward.nextcloud my-remote:default/nextcloud/2001:db8::1
lxd_network_forward.nextcloud: Importing from ID "my-remote:default/nextcloud/2001:db8::1"...
lxd_network_forward.nextcloud: Import prepared!
  Prepared lxd_network_forward for import
lxd_network_forward.nextcloud: Refreshing state...
data.lxd_instance.nextcloud: Reading...
data.lxd_instance.nextcloud: Read complete after 0s [name=nextcloud]

Import successful!
```

### LXD version

6.5

v2.6.0

### Terraform/Tofu version

OpenTofu v1.10.6

Ubuntu 25.04

```hcl
# I believe this to be a minimal relevant reproducer. It could
# probably be simplified more by using bridged network, instead of OVN.

resource "lxd_network_forward" "nextcloud" {
  network        = lxd_network.nextcloud.name
  listen_address = "2001:db8::1"


  ports = [
    {
      description    = "SSH"
      protocol       = "tcp"
      listen_port    = "22"
      target_port    = "22"
      target_address = "fd::1"
    },
  ]
}

resource "lxd_network" "uplink" {
  name        = "uplink"
  type        = "physical"
  description = "OVN external connection"

  config = {
    "ipv6.gateway"    = "fe80::2/64"
    "ipv6.ovn.ranges" = "2001:db8::ff:10-2001:db8::ff:20"
    "ipv6.routes"     = "2001:db8::/64"
    "parent"          = "enp3s0"
  }
}

resource "lxd_network" "nextcloud" {
  name = "nextcloud"
  type = "ovn"

  config = {
    "bridge.mtu"                    = "1442"
    "ipv4.nat"                      = "false"
    "ipv6.address"                  = "fd::ff:1/64"
    "ipv6.nat"                      = "true"
    "network"                       = lxd_network.uplink.name 
    "volatile.network.ipv6.address" = "2001:db8::ff:11"
  }
}
```

### Terraform output

Included in the description.

MusicDin:
Closes #628

### DIFF
--- a/internal/common/lxd_import_test.go
+++ b/internal/common/lxd_import_test.go
@@ -86,6 +86,14 @@ func TestSplitImportID_basic(t *testing.T) {
 			Result:   map[string]string{"name": "vm", "project": "proj", "remote": "rem"},
 		},
 		{
+			ImportID: "/1000:aaaa::200",
+			Result:   map[string]string{"name": "1000:aaaa::200"},
+		},
+		{
+			ImportID: "rem:/1000:aaaa::200",
+			Result:   map[string]string{"remote": "rem", "name": "1000:aaaa::200"},
+		},
+		{
 			ImportID:    "",
 			ErrorString: "Import ID cannot be empty.",
 		},


internal/common/lxd_import.go
```diff
@@ -71,9 +71,12 @@ func (m ImportMetadata) ParseImportID(importID string) (map[string]string, diag.
 func processFields(id string, requiredFields []string) (map[string]string, error) {
 	result := make(map[string]string)
 
-	// Split id into [remote:]<id>
+	// Check for remote if import ID contains colon.
+	// If colon appears after slash, it is not a remote (most likely part of the IPv6).
+	colonIndex := strings.Index(id, ":")
+	slashIndex := strings.Index(id, "/")
```
simondeziel: Feels like a good use case for `before, after, found := strings.Cut(id, '/')`.

simondeziel (approved):
Thanks

Issue #637, opened 2025-11-24
Changes to file content in Terraform files do not result in changes to existing instance


Vultaire:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Issue description

It appears that updates to the files clauses in lxd_instances don't actually result in file modifications, despite Terraform's output suggesting that it might.  I tried doing this to update the definition of an systemd config file for a custom server, but found out that it did not work.

### LXD version

5.21.4-8a3cf61

### LXD Terraform provider version

2.6.1

### Terraform/Tofu version

1.12.2

### Operating system / distribution

Ubuntu 24.04

### Relevant Terraform configuration

```hcl
# **TWO STEP PROCESS REQUIRED** to reproduce this.
# We need to apply an initial bundle, and then a change.

# === INITIAL BUNDLE ===

terraform {
  required_providers {
    lxd = {
      source = "terraform-lxd/lxd"
      version = "2.6.1"
    }
  }
}
provider "lxd" {}
resource "lxd_instance" "file_update_test" {
  name = "file-update-test"
  image = "ubuntu:24.04"
  file {
    content = "foo"
    target_path = "/root/test.txt"
  }
}


# ----------------------------------------------------------------------
# === UPDATED BUNDLE ===

terraform {
  required_providers {
    lxd = {
      source = "terraform-lxd/lxd"
      version = "2.6.1"
    }
  }
}
provider "lxd" {}
resource "lxd_instance" "file_update_test" {
  name = "file-update-test"
  image = "ubuntu:24.04"
  file {
    content = "bar"
    target_path = "/root/test.txt"
  }
}

# ----------------------------------------------------------------------
# "lxc exec file-update-test cat /root/test.txt" should output "bar" after applying
# the second bundle, but it will output "foo" since the file has not been changed.
# This is despite Terraform's output claiming that the file would be changed.
```

### Terraform output

For the initial apply:

```
$ terraform apply -auto-approve

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # lxd_instance.file_update_test will be created
  + resource "lxd_instance" "file_update_test" {
      + allow_restart    = false
      + config           = {}
      + description      = ""
      + ephemeral        = false
      + image            = "ubuntu:24.04"
      + interfaces       = (known after apply)
      + ipv4_address     = (known after apply)
      + ipv6_address     = (known after apply)
[...]

PR #641, opened 2025-12-11
instance: Fix file upload when target path does not change

MusicDin:
Fixes: https://github.com/terraform-lxd/terraform-provider-lxd/issues/637


nmezhenskyi (approved):
Thanks!

PR #709, opened 2026-04-23
Code quality

simondeziel:
Depends on https://github.com/canonical/lxd/pull/18183

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,10 +70,10 @@ jobs:
 
       - name: Download minio/mc
         uses: canonical/lxd/.github/actions/download-minio@main # zizmor: ignore[unpinned-uses]
-        if: ${{ matrix.channel != '5.0/stable' && matrix.channel != '5.0/edge' }}
+        if: ${{ startsWith(matrix.channel, '5.21/') || matrix.channel == '6/stable' }}
 
       - name: Configure MinIO
-        if: ${{ matrix.channel != '5.0/stable' && matrix.channel != '5.0/edge' }}
+        if: ${{ startsWith(matrix.channel, '5.21/') || matrix.channel == '6/stable' }}
         run: |
           mkdir -p /opt/minio
           mv "$(go env GOPATH)/bin/minio" /opt/minio/
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,7 +69,7 @@ jobs:
           lxc remote add localhost "$(lxc config trust add --name lxd-terraform-provider --quiet)"
 
       - name: Download minio/mc
-        uses: canonical/lxd/.github/actions/download-minio@main # zizmor: ignore[unpinned-uses]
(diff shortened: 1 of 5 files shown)


internal/provider-config/config.go
```diff
@@ -474,7 +474,14 @@ func DetermineLXDAddress(protocol string, address string) (string, error) {
 				port = "443"
 			}
 
-			url.Host = url.Hostname() + ":" + port
+			host := url.Hostname()
+			// If the hostname contains ':' it's likely an IPv6 address and
+			// must be enclosed in brackets when adding a port.
+			if strings.Contains(host, ":") {
+				url.Host = "[" + host + "]:" + port
```
MusicDin, line 481: Does url.Hostname() parse IPv6 without brackets, or do we need to check for the presence of brackets as well?

internal/common/lxd_exec.go
```diff
@@ -86,8 +86,8 @@ func (e *ExecModel) Execute(ctx context.Context, server lxd.InstanceServer, inst
 		Interactive:  false,
 		RecordOutput: false,
 		Cwd:          e.WorkingDir.ValueString(),
-		User:         uint32(e.GroupID.ValueInt64()),
-		Group:        uint32(e.UserID.ValueInt64()),
+		User:         uint32(e.UserID.ValueInt64()),
+		Group:        uint32(e.GroupID.ValueInt64()),
```
MusicDin, line 90: Aj jaj jaj 🫤 Nice catch 🚀

internal/provider-config/config.go
simondeziel: https://pkg.go.dev/net/url#URL.Hostname says:
> If the result is enclosed in square brackets, as literal IPv6 addresses are, the square brackets are removed from the result.

.github/workflows/test.yml
```diff
@@ -70,10 +70,10 @@ jobs:
 
       - name: Download minio/mc
         uses: canonical/lxd/.github/actions/download-minio@main # zizmor: ignore[unpinned-uses]
-        if: ${{ matrix.channel != '5.0/stable' && matrix.channel != '5.0/edge' }}
+        if: ${{ startsWith(matrix.channel, '5.21/') || matrix.channel == '6/stable' }}
 
       - name: Configure MinIO
-        if: ${{ matrix.channel != '5.0/stable' && matrix.channel != '5.0/edge' }}
```
MusicDin, line 76: Do you know if LXD 5.21 and 6 still have API extension for local buckets? 
We are checking for presence of `storage_buckets_local` API ext., but tests still run.

https://github.com/terraform-lxd/terraform-provider-lxd/blob/2e8592d49c5abfe528f21630e55edf8c63a9dd78/internal/storage/resource_storage_bucket_test.go#L20

simondeziel: AFAIK, this is not an API problem.

The current CI breakage is due to the CI action now downloading `mc` properly.

nmezhenskyi (approved):
LGTM

PR #721, opened 2026-05-30
Apply code quality suggestion

simondeziel:
Copilot suggested extending the list of accepted DNS record types but this is pointless IMHO as LXD supports ~all valid types (those supported by `github.com/miekg/dns`).

### DIFF
--- a/internal/network/resource_network_zone_record.go
+++ b/internal/network/resource_network_zone_record.go
@@ -119,7 +119,7 @@ func (r NetworkZoneRecordResource) Schema(_ context.Context, _ resource.SchemaRe
 							Computed:    true,
 							Description: "Record entry TTL",
 							Validators: []validator.Int64{
-								int64validator.AtLeast(1),
+								int64validator.AtLeast(0),
 							},
 						},
 					},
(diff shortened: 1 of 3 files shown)


internal/network/resource_network_zone_record.go
```diff
@@ -103,11 +103,9 @@ func (r NetworkZoneRecordResource) Schema(_ context.Context, _ resource.SchemaRe
 					Attributes: map[string]schema.Attribute{
 						"type": schema.StringAttribute{
 							Required:    true,
-							Description: "Record entry type",
+							Description: "Record entry type. All valid DNS types are accepted (e.g. A, AAAA, CNAME, MX, TXT).",
 							Validators: []validator.String{
-								stringvalidator.OneOf(
-									"A", "AAAA", "CNAME", "TXT",
-								),
+								stringvalidator.LengthAtLeast(1),
```
MusicDin, line 108: Can we add the remaining ones to the oneOf validator instead?
The benefit is that this influences the schema. If typo is made, it will produce an intuitive error during planning (and in an editor, if respective extension is available).

simondeziel: The list of supported value is quite big (https://en.wikipedia.org/wiki/List_of_DNS_record_types gives just an overview) and keeps changing (although infrequently but still).

MusicDin (approved):

PR #730, opened 2026-06-09
storage: Support for Alletra, Powerflex, Powerstore, and PureStorage storage drivers

MusicDin:
Adds support for storage drivers `alletra`, `powerflex`, `powerstore`, and `pure`.

### DIFF
--- a/docs/resources/storage_pool.md
+++ b/docs/resources/storage_pool.md
@@ -68,7 +68,7 @@ for more details on how to create a storage pool in clustered mode.
 
 * `name` - **Required** - Name of the storage pool.
 
-* `driver` - **Required** - Storage pool driver. Must be one of `dir`, `zfs`, `lvm`, `btrfs`, `ceph`, `cephfs`, or `cephobject`.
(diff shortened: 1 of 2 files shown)


MusicDin:
@roosterfish I went ahead and added `powerflex.version` as computed key (besides `powerflex.mode` which is set if user does not provide one). Is there any other PowerFlex storage pool key that LXD might set by itself?

simondeziel (approved):
I'm ambivalent on having the driver lists sorted or not. Some of them are so specialized that they are probably fine to not appear early on.

roosterfish:
> Is there any other PowerFlex storage pool key that LXD might set by itself?

The only other one is `powerflex.user.name` which defaults to `admin` if not set.

MusicDin:
> > Is there any other PowerFlex storage pool key that LXD might set by itself?
> 
> The only other one is `powerflex.user.name` which defaults to `admin` if not set.

Thanks, will add `powerflex.user.name` as well, and check if similar is true for Alletra.

PR #732, opened 2026-06-10
storage: Add PowerStore and PowerFlex username as a computed key

MusicDin:


simondeziel (dismissed):

internal/storage/resource_storage_pool.go
```diff
@@ -775,10 +775,12 @@ func (m StoragePoolModel) ComputedKeys(driver string) []string {
 		keys = []string{
 			"powerflex.mode",
 			"powerflex.version",
```
roosterfish, line 777: `volatile.powerflex.version`.

roosterfish (changes requested):

internal/storage/resource_storage_pool.go
MusicDin: Have removed `powerflex.version`, as `volatile.*` keys are considered computed by default

simondeziel (approved):

PR #734, opened 2026-06-12
network/test: Fix flaky test by waiting for network

MusicDin:
Network attach from profile test is flaky because it expects ipv4/6 to be set, but does not wait for one.


simondeziel (approved):
Ah, thanks!!

PR #745, opened 2026-06-25
internal: Error out if resource is not found after resource creation or update

MusicDin:
If resource is read and not found on the remote server (removed outside Terraform), remove its state so that Terraform considers that as missing resource and plans creation of the resource.

If resource is being created or updated and is not found afterwards, retain the state and error out. This indicates bigger underlying issue which is currently silently ignored (by state being removed) and results in Terraform panic due to missing state.

Also taint all resources upon creation, so if sync fails, the resource is guaranteed to be tracked.

Closes #701


simondeziel (approved):
Nice, thanks

PR #750, opened 2026-06-29
instance: Allow container creation with unknown image

MusicDin:
This PR fixes regression with unknown container image value, which is treated as empty and fails the validation requiring container image to have an image provided.

Reported by @roosterfish

### DIFF
--- a/internal/instance/resource_instance.go
+++ b/internal/instance/resource_instance.go
@@ -572,7 +572,7 @@ func (r InstanceResource) ValidateConfig(ctx context.Context, req resource.Valid
 	}
 
 	// Ensure empty container cannot be started.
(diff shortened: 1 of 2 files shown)


roosterfish (approved):
Thanks for the quick fix! 

Out of curiosity, why is the image considered unknown in this case, due to the `for_each` loop?

MusicDin:
> Out of curiosity, why is the image considered unknown in this case, due to the for_each loop?

It is because fingerprint of a `lxd_image` resource is a computed field known after terraform apply. During the plan it is not known because image and its metadata was not fetched yet at that point.

roosterfish:
> It is because fingerprint of a `lxd_image` resource is a computed field known after terraform apply. During the plan it is not known because image and its metadata was not fetched yet at that point.

I see, thanks!

PR #753, opened 2026-07-01
Add Network ACL datasource

tugbataluy:
Closes #677.

Adds a `lxd_network_acl` data source so users can look up an existing network ACL by name and use its attributes (description, config, ingress/egress rules) elsewhere in their configuration.

```hcl
data "lxd_network_acl" "acl" {
  name = "my-acl"
}
```

Supports optional `project` and `remote` arguments, consistent with other data sources in this provider.

### DIFF
--- a/docs/data-sources/network_acl.md
+++ b/docs/data-sources/network_acl.md
@@ -1,0 +1,41 @@
+# lxd_network_acl
+
+Provides information about an existing LXD network ACL.
+
+## Example Usage
+
+```hcl
+data "lxd_network_acl" "acl" {
+  name = "my-acl"
+}
+```
+
+## Argument Reference
+
+* `name` - **Required** - Name of the network ACL.
+
+* `project` - *Optional* - Name of the project where the ACL is located.
+
+* `remote` - *Optional* - The remote in which the resource was created. If
+  not provided, the provider's default remote is used.
+
+## Attribute Reference
+
+This data source exports the following attributes in addition to the arguments above:
+
+* `description` - Description of the network ACL.
+
+* `config` - Map of key/value pairs of network ACL config settings.
+
+* `egress` - Set of egress rules. Each rule exports:
+  * `action` - Rule action (`allow`, `drop`, `reject`).
+  * `state` - Rule state (`enabled`, `disabled`, `logged`).
+  * `description` - Rule description.
+  * `source` - Source address/CIDR.
+  * `destination` - Destination address/CIDR.
+  * `destination_port` - Destination port(s).
+  * `protocol` - Protocol (`tcp`, `udp`, `icmp4`, `icmp6`).
+  * `icmp_type` - ICMP type (only relevant for icmp4/icmp6).
+  * `icmp_code` - ICMP code (only relevant for icmp4/icmp6).
+
+* `ingress` - Set of ingress rules. Same attributes as `egress`.
--- a/internal/network/datasource_network_acl.go
+++ b/internal/network/datasource_network_acl.go
@@ -1,0 +1,157 @@
+package network
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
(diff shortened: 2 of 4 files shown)


docs/data-sources/network_acl.md
```diff
@@ -0,0 +1,41 @@
...
+## Attribute Reference
+
+This data source exports the following attributes in addition to the arguments above:
+
+* `description` - Description of the network ACL.
+
+* `config` - Map of key/value pairs of network ACL config settings.
+
+* `egress` - Set of egress rules. Each rule exports:
+  * `action` - Rule action (`allow`, `allow-stateless`, `drop`, `reject`).
```
simondeziel: `s/stateless/related/`? Because `allow-stateless` doesn't seem supported by LXD yet: https://github.com/canonical/lxd/blob/be2dd46998d4a2d5f4890b7a13e853c864a90c05/lxd/network/acl/acl_ovn.go#L445-L455

internal/network/datasource_network_acl.go
```diff
@@ -0,0 +1,157 @@
...
+	state.Egress = egress
+	state.Ingress = ingress
+
+	diags = resp.State.Set(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+}
+
+// aclRuleAttributesComputed returns ACL rule attributes with all fields Computed,
+// for use in data source schemas where no validators or defaults are needed.
+func aclRuleAttributesComputed() map[string]schema.Attribute {
```
MusicDin: Can we name that `aclRuleAttributes`?

internal/network/datasource_network_acl_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctest.PreCheck(t) },
+		ProtoV6ProviderFactories: acctest.ProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: acctest.Provider() + testAccNetworkACL_DS_withRules(aclName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.lxd_network_acl.acl", "name", aclName),
+					resource.TestCheckResourceAttr("data.lxd_network_acl.acl", "egress.#", "1"),
```
MusicDin, line 47: Can you ensure the resulting egress fields have expected values?

Also, one test for ingress would be appreciated.

MusicDin (changes requested):

internal/network/datasource_network_acl.go
tugbataluy: There's already an `aclRuleAttributes()` function in `resource_network_acl.go` that returns `resource/schema.Attribute` with validators and Required/Optional fields. Since the datasource version returns `datasource/schema.Attribute` (all Computed, no validators), they can't share the same name. I've renamed it to `aclDatasourceRuleAttributes()` to keep it clear which schema type it serves.

docs/data-sources/network_acl.md
tugbataluy: Good catch, thanks @simondeziel ! LXD currently only supports `allow`, `drop`, and `reject` as user-facing actions (allow is internally mapped to allow-related at the OVN level). Removed allow-stateless from the datasource docs

internal/network/datasource_network_acl_test.go
tugbataluy: Done. The `TestAccNetworkACL_DS_withRules` test now uses TestCheckTypeSetElemNestedAttrs to verify the egress rule's field values (action, destination, protocol, state). Also added a new `TestAccNetworkACL_DS_withIngressRules` test that creates an ACL with an ingress rule and asserts the expected field values.

internal/network/datasource_network_acl.go
MusicDin: Perfect, thanks.

MusicDin (approved):

simondeziel (approved):
Thanks!

PR #762, opened 2026-07-31
auth: Allow creating pending TLS identities

### LINKED ISSUE #760: Regression in LXD 5.21.5+: lxd_trust_token replacement fails due to lingering expanded identities.
### Please confirm

### Issue description

When an `lxd_trust_token` resource is created and subsequently consumed by a client (e.g., Juju or another remote lxc client), LXD expands the pending token into an active, permanent client certificate identity in LXD's trust database (lxc config trust / lxc auth identity).

When Terraform replaces or destroys the `lxd_trust_token` resource, the provider revokes or removes the token object, but the expanded active certificate identity matching the token's name remains in LXD's database.

Starting in LXD 5.21.5, strict identity checks prevent clients from completing a handshake or registering a new client certificate under an identity name that is already bound to an active certificate. As a result, when a replaced token with the same name is delivered to a client, identity finalization fails:

```bash
This "identity" entry already exists
```

I tried to simplify my reproducer (I hope it is accurate). The full reproducer is a CI that runs an LXD preconfigure module and then a Terragrunt stack to bootstrap Juju, then use it to install MAAS. Here is a link to a failed run: https://github.com/canonical/maas-terraform-modules/actions/runs/30411398121

### LXD version

5.21.5

2.7.1

### Terraform/Tofu version

1.15.8

Ubuntu 24.04

```hcl
terraform {
  required_providers {
    lxd = {
      source  = "terraform-lxd/lxd"
      version = "~> 2.6"
    }
  }
}

resource "lxd_trust_token" "maas_charms" {
  name = "maas-charms"
}



terraform apply
lxc remote add _token_
terraform apply -replace=lxd_trust_token.maas_charms
lxc remote remove _old_remote_
lxc remote add _new_token_
```

### Terraform output

N/A

MusicDin:
The `lxd_auth_identity` field `tls_certificate` is now optional. If it is not set, LXD creates a pending identity and returns a trust_token (set as computed field). Other client can use the trust token to enroll its own certificate.

The trust token is stored in state because once the identity is created, LXD cannot reissue it. If it expires before it is redeemed, the identity is replaced. It is also replaced if `tls_certificate` is set while the identity is pending.

Once the token is consumed, the identity becomes a regular TLS identity. There is no plan diff, and later changes, including a new certificate, are updated in place.

Requires the `access_management_tls` API extension.

### DIFF
--- a/docs/resources/auth_identity.md
+++ b/docs/resources/auth_identity.md
@@ -21,6 +21,29 @@ resource "lxd_auth_identity" "tls-identity" {
 }
 ```
 
+### Pending TLS identity
+
+If `tls_certificate` is omitted for a `tls` identity, LXD creates the identity in a
+pending state and issues a trust token. The token is handed to an untrusted client,
+which redeems it to enroll its own certificate:
+
+```hcl
+resource "lxd_auth_identity" "pending-identity" {
+  auth_method = "tls"
+  name        = "jane"
+  groups      = ["admins"]
+}
+
+output "trust_token" {
+  value     = lxd_auth_identity.pending-identity.trust_token
+  sensitive = true
+}
+```
+
+The client then runs `lxc remote add <remote> <token>`.
+
+Requires the `access_management_tls` API extension.
+
 ## Argument Reference
 
 * `name` - **Required** - Name of the identity.
@@ -29,10 +52,41 @@ resource "lxd_auth_identity" "tls-identity" {
 
 * `groups` - *Optional* - List of group names to add this identity to.
 
-* `tls_certificate` - *Optional* - PEM encoded x509 certificate. Must be set when authentication method is `tls`.
+* `tls_certificate` - *Optional* - PEM encoded x509 certificate. Applicable only when the
+  authentication method is `tls`. If omitted, a pending TLS identity is created and a trust
+  token is issued instead. Once a TLS identity is created, removing this attribute from the
+  configuration leaves the certificate intact.
 
-* `remote` - *Optional* - The remote in which the resource will be created. If
-	not provided, the provider's default remote will be used.
+* `remote` - *Optional* - The remote in which the resource will be created. If not provided,
+  the provider's default remote will be used.
+
+## Attribute Reference
+
+This resource exports the following attributes in addition to the arguments above:
+
+* `trust_token` - Trust token that an untrusted client uses to enroll its certificate. It is
+  set only for a pending TLS identity, and is cleared once the token has been redeemed.
+
+* `expires_at` - Time at which the trust token expires, in RFC 3339 format and UTC. It is null when
+  the server config `core.remote_token_expiry` is not set, which means the token does not expire.
+
+## Pending TLS trust token lifecycle
+
+A trust token is returned only when a pending TLS identity is created, which means no
+`tls_certificate` was set, and can never be retrieved from the server afterwards. LXD
+provides no way to reissue a token for an existing identity, so if the token expires before
+it is redeemed, the identity is unusable and Terraform plans a replacement of the resource.
+
+Setting `tls_certificate` on a TLS identity whose token is still outstanding also plans a
+replacement. LXD does not allow a certificate to be set on a pending identity, and the
+outstanding token is invalidated along with the identity it was issued for.
+
+Once the token is redeemed, the identity holds a certificate and is never replaced on that
+account. Subsequent changes, such as group membership or a new `tls_certificate`, are
+applied in place.
+
+An imported pending TLS identity has no `trust_token`, because the token cannot be read back
+from the server.
 
 ## Importing
 
--- a/internal/acctest/checks.go
+++ b/internal/acctest/checks.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	lxd "github.com/canonical/lxd/client"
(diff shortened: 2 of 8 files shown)


internal/acctest/checks.go
```diff
@@ -177,15 +178,23 @@ func PreCheckRoot(t *testing.T) {
...
 
-// PreCheckLocalServerHTTPS skips the test if the server is not exposed on the localhost
-// over port 8443. This is required for remote provider tests.
-func PreCheckLocalServerHTTPS(t *testing.T) {
-	conn, err := net.DialTimeout("tcp", "127.0.0.1:8443", 1*time.Second)
+// PreCheckLocalServerHTTPS skips the test if the LXD server is not reachable over HTTPS.
+// Additionally, it returns the server's HTTPS address. If the test remote is a unix
+// socket, the local HTTPS address is used instead.
+func PreCheckLocalServerHTTPS(t *testing.T) string {
+	address := testRemotes()[testProviderRemoteName].Address
```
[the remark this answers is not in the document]
MusicDin: This is incorrect. Go does allow discarding function's return values. And Prechecks are specific in that they skip the test if condition is not met anyway, but in some cases returns optional useful values, such as address in this case.

internal/auth/resource_identity.go
```diff
@@ -332,3 +472,54 @@ func (r *AuthIdentityResource) ImportState(ctx context.Context, req resource.Imp
...
+	if expiresAt.IsNull() || expiresAt.IsUnknown() {
+		return false
+	}
+
+	parsed, err := time.Parse(time.RFC3339, expiresAt.ValueString())
+	if err != nil {
+		return false
+	}
+
+	return now.After(parsed)
```
simondeziel, line 509: To avoid TZ issues:
```suggestion
	return now.UTC().After(parsed.UTC())
```

simondeziel (changes requested):

internal/auth/resource_identity.go
MusicDin: Thanks, I have fixed it per you suggestion where expiry is returned as string (in `tokenExpiry` func).

In comparison itself it is not needed. The Go explicitly states that timezone is only used for representation. Comparison is always standardized.

internal/auth/resource_identity.go
```diff
@@ -261,6 +341,50 @@ func (r AuthIdentityResource) Delete(ctx context.Context, req resource.DeleteReq
...
+	}
+
+	// A pending identity is the only TLS identity that LXD reports without a certificate.
+	isPending := state.AuthMethod.ValueString() == "tls" && state.Certificate.IsNull()
+
+	// The attribute that is reported as forcing the replacement.
+	var replaceOn path.Path
+
+	switch {
+	case isTrustTokenExpired(isPending, state.ExpiresAt, time.Now()):
```
tugbataluy, line 371: When the token expires, the plan shows `trust_token: (sensitive value) -> (known after apply) # forces replacement`. That tells the reader an identity is about to be destroyed but not why. The
expiry is a wall-clock decision the user did not make, so I think we can explain it with a sentence.

```suggestion
	case isTrustTokenExpired(isPending, state.ExpiresAt, time.Now()):
		replaceOn = path.Root("trust_token")
		resp.Diagnostics.AddWarning(
			fmt.Sprintf("Trust token for %q identity %q has expired", state.AuthMethod.ValueString(), state.Name.ValueString()),
			"The identity is still pending and LXD cannot reissue a trust token, so the identity will be replaced and a new token issued.",
		)
```

docs/resources/auth_identity.md
```diff
@@ -29,10 +52,40 @@ resource "lxd_auth_identity" "tls-identity" {
 
 * `groups` - *Optional* - List of group names to add this identity to.
 
-* `tls_certificate` - *Optional* - PEM encoded x509 certificate. Must be set when authentication method is `tls`.
+* `tls_certificate` - *Optional* - PEM encoded x509 certificate. Applicable only when the
```
tugbataluy, line 55: `tls_certificate` moved from `Optional` to `Optional + Computed` ([schema](https://github.com/terraform-lxd/terraform-provider-lxd/blob/0c4ce6ccaf8a57ffcbeff2dae675da9d4190f601/internal/auth/resource_identity.go#L82-L89)).

That changes behaviour for people who already use this resource: removing the attribute from config used to clear the certificate on the server, and now it is a no-op, because [`Update`](https://github.com/terraform-lxd/terraform-provider-lxd/blob/0c4ce6ccaf8a57ffcbeff2dae675da9d4190f601/internal/auth/resource_identity.go#L291-L296) carries the server value forward. The argument entry only covers the create path.

```suggestion
* `tls_certificate` - *Optional* - PEM encoded x509 certificate. Applicable only when the  authentication method is `tls`. If omitted, a pending TLS identity is created and a trust token is issued instead. Once an identity holds a certificate, removing this attribute from the configuration leaves the certificate in place rather than clearing it.
```

internal/auth/resource_identity.go
MusicDin: Fixed

docs/resources/auth_identity.md
MusicDin: Fixed

simondeziel (approved):
Thx

PR #770, opened 2026-08-28
auth: Client bearer tokens

MusicDin:
Adds `lxd_auth_identity_token` resource that allows generating client bearer tokens.

The identity resource is also changed to use `type` instead of `auth_method`. Type is expected to be short type format, such as `tls` or `bearer`.

To prevent breaking changes, exactly one of the fields `auth_method` and `type` must be set. The warning is emitted when `auth_method` is used and changing configuration input from `auth_method` to `type` is noop because Terraform will now populated both fields. 
In the future releases, `auth_method` will become read-only (computed) and `type` required. I've reflected that in the documentation already to reduce confusion and added note at the end.

Requires rebase after: 
- https://github.com/terraform-lxd/terraform-provider-lxd/pull/762


docs/resources/auth_identity_token.md
```diff
@@ -0,0 +1,77 @@
...
+
+The `lxd_auth_identity_token` resource allows you to issue a token for an LXD
+identity of type `bearer`.
+
+## Required API extensions
+
+The target LXD server must support the following API extensions:
+
+* `access_management_expiry`
+* `auth_bearer`
```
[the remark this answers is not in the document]
MusicDin: Those are different extensions. Client bearer tokens require `auth_bearer` extension

MusicDin:
@simondeziel Ahh thanks, accidentally added assignees instead of reviewers

internal/auth/resource_identity_test.go
```diff
@@ -264,8 +385,6 @@ func testAccIdentity_bearer(name string, groups []string) string {
 	)
 }
 
-// testAccIdentity_captureAttr stores the value of the given resource attribute in target, so that
```
tugbataluy, line 267: Looks like that two lined comment got lost in the move, I think we need to keep that comment since it's the only helper in the file that mutates a pointer across steps.

```suggestion
// testAccIdentity_captureAttr stores the value of the given resource attribute in target, so that
// it can be used by a subsequent test step.
```

internal/auth/resource_identity_test.go
```diff
@@ -284,6 +403,71 @@ func testAccIdentity_captureAttr(resourceName string, attrName string, target *s
...
+                }
+        `,
+		identityType,
+		name,
+		acctest.QuoteStrings(groups),
+	)
+}
+
+// testAccIdentityCert is a self signed certificate for TLS identities.
+
```
tugbataluy: Blank line between the comment and the `const` means godoc won't attach it.
```suggestion
// testAccIdentityCert is a self signed certificate for TLS identities.
const testAccIdentityCert = `
```

internal/auth/resource_identity_test.go
```diff
@@ -313,34 +497,39 @@ func testAccIdentity_tlsPendingWithCertificate(name string, groups []string, cer
...
+		name,
+	)
+}
 
-		  allowed_uses = [
-		    "digital_signature"
-		  ]
-		}
+// testAccIdentity_bearerWithCertificate is rejected before the identity is
+// created, so the certificate does not have to be a valid one.
```
tugbataluy: Same as above:

```suggestion
// testAccIdentity_bearerWithCertificate is rejected before the identity is
// created, so the certificate does not have to be a valid one.
func testAccIdentity_bearerWithCertificate(name string) string {
```

internal/auth/resource_identity_internal_test.go
```diff
@@ -109,3 +109,30 @@ func TestIsCertificateSetOnPending(t *testing.T) {
 		})
 	}
 }
+
+func TestExpiryPattern(t *testing.T) {
```
tugbataluy: Just an idea:
This tests `expiryPattern` / `expiryNonZeroPattern` from `resource_identity_token.go`. Would you mind moving it to a new `resource_identity_token_internal_test.go`? Keeps the 1:1 file pairing the package already follows.

internal/auth/resource_identity_token.go
```diff
@@ -0,0 +1,304 @@
...
+
+		resp.Diagnostics.AddError(fmt.Sprintf("Failed to retrieve the token expiry of identity %q", identityName), err.Error())
+		return
+	}
+
+	// The identity reports the expiry of the token it currently bears. If no
+	// expiry is reported, or the reported expiry differs from the recorded one,
+	// the token in the state has been revoked or replaced. An expired token is
+	// dropped as well, because the server keeps reporting its expiry.
+	if expiresAt != state.ExpiresAt || isExpired(expiresAt) {
```
tugbataluy: Works today because both fields are comparable, but `Equal` is the framework's contract and reads as intent.

```suggestion
	if !expiresAt.Equal(state.ExpiresAt) || isExpired(expiresAt) {
```

internal/auth/resource_identity_token.go
```diff
@@ -0,0 +1,304 @@
...
+		if errors.IsNotFoundError(err) {
+			// The identity, and with it its token, is already gone.
+			return
+		}
+
+		resp.Diagnostics.AddError(fmt.Sprintf("Failed to retrieve the token expiry of identity %q", identityName), err.Error())
+		return
+	}
+
+	if expiresAt != state.ExpiresAt {
```
tugbataluy: Same reasoning above.
```suggestion
	if !expiresAt.Equal(state.ExpiresAt) {
```

docs/resources/auth_identity.md
```diff
@@ -48,7 +48,7 @@ Requires the `access_management_tls` API extension.
 
 * `name` - **Required** - Name of the identity.
 
-* `auth_method` - **Required** - Authentication method, can be either `tls` or `bearer`.
+* `type` - **Required** - Identity type, can be `tls` or `bearer`.
```
tugbataluy, line 51: Documenting `type` as **Required** now is a fair call (matches the PR description). A pointer to the note at the bottom saves readers who still use `auth_method` from a "wait, why isn't this erroring?" moment.

```suggestion
* `type` - **Required** - Identity type, can be `tls` or `bearer`. See [Note on `auth_method`](#note-on-auth_method).
```

What would you think?

tugbataluy:
Really nice work on this @MusicDin :sparkle: .  The migration from `auth_method` to `type` is handled carefully, the tainted-state path on token issue is solid. Left a few tiny nits, none blocking.

internal/auth/resource_identity_test.go
MusicDin: Fixed

simondeziel (approved):

tugbataluy (approved):

nmezhenskyi (approved):

PR #778, opened 2026-09-08
storage|network: Match metadata configuration key placeholders

### LINKED ISSUE #777: Cannot set `bgp.peers.foo.asn` for a physical network
### Please confirm

### Issue description

Terraform gives me the following error : 
```
Network "UPLINK" (physical) does not support config key "bgp.peers.foo.asn"
```

@MusicDin said "When comparing keys to metadata, placeholders are not resolved, so `bgp.peers.NAME.asn` is compared to `bgp.peers.foo.asn` as exact match and fails."

Thanks !

### LXD version

6/stable

Commit c0f202a

### Terraform/Tofu version

1.16.1

_No response_

```hcl

```

### Terraform output

_No response_

MusicDin:
When the metadata API lists the available config keys, some keys contain placeholders, for example `bgp.peers.NAME.asn` or `user.*`. The provider compared config keys with these entries verbatim, so any key with a placeholder was rejected as unsupported.

This PR deduplicates shared logic between network and storage pool per-member configuration and adds a shared matcher in `internal/common` that resolves placeholders by the following rules:

1. A trailing `*` matches the rest of the key (`user.*`).
2. A placeholder wrapped in `{}` or `<>` matches any non-empty string, dots included (`storage.project.{name}.backups_volume`, project `my.proj` -> `storage.project.my.proj.backups_volume`).
3. An uppercase name matches exactly one segment (`bgp.peers.NAME.asn` -> Placeholder value does not allow dots).
4. Any other segment is matched literally.

### DIFF
--- a/internal/common/lxd_metadata.go
+++ b/internal/common/lxd_metadata.go
@@ -2,6 +2,8 @@ package common
 
 import (
 	"fmt"
+	"regexp"
+	"strings"
 	"sync"
 
 	lxd "github.com/canonical/lxd/client"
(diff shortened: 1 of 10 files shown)


internal/common/lxd_metadata.go
```diff
@@ -2,19 +2,164 @@ package common
...
+	Keys   api.MetadataConfigurationConfigKeys
+	Prefix string
+}
+
+// MetadataConfigKeys matches concrete configuration keys to their metadata.
+type MetadataConfigKeys struct {
+	// Map of exact keys.
+	exact map[string]api.MetadataConfigurationConfigKey
+
+	// Map of keys with dynamic segments, such as wildcards or named placeholders.
```
tugbataluy, line 26: `patterns` is a slice, the comment says map.

```suggestion
	// List of keys with dynamic segments, such as wildcards or named placeholders.
```

internal/storage/resource_storage_pool.go
```diff
@@ -612,15 +554,16 @@ func (m StoragePoolModel) ParsePoolConfigs(ctx context.Context, server lxd.Insta
...
 	}
 
-	if len(allPoolKeys) > 0 {
-		for k := range poolConfig {
-			if !slices.Contains(allPoolKeys, k) {
-				return nil, nil, fmt.Errorf("Storage pool %q (%s) does not support config key %q", poolName, driver, k)
+	if server.CheckExtension("metadata_configuration") == nil {
+		for key := range poolConfig {
+			_, ok := configKeys.Lookup(key)
+			if !ok && !strings.HasPrefix(key, "user.") {
```
tugbataluy, line 565: LXD accepts `user.*` on storage pools (`lxd/storage/drivers/driver_common.go`, "User keys are not validated") but `pool-conf` metadata does not list `user.*` for any driver, so the matcher alone would reject it. The bypass is load-bearing here and `TestAccStoragePool_userConfig` depends on it. A one-liner stops someone "cleaning it up" later.

```suggestion
			// LXD accepts "user.*" keys on storage pools, but does not list them in the pool metadata.
			_, ok := configKeys.Lookup(key)
			if !ok && !strings.HasPrefix(key, "user.") {
```

internal/network/resource_network_test.go
```diff
@@ -47,6 +47,47 @@ func TestAccNetwork_basic(t *testing.T) {
...
+				}, nil),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("lxd_network.network", "config.bgp.peers.peer1.address", "192.0.2.1"),
+					resource.TestCheckResourceAttr("lxd_network.network", "config.bgp.peers.peer1.asn", "65000"),
+				),
+			},
+		},
+	})
+}
+
```
tugbataluy, line 90: Both new tests cover keys that are now accepted. Nothing pins that unknown keys are still rejected, and this PR replaces that logic. Maybe having  a negative test case would be better. What do you think?

```suggestion
}

func TestAccNetwork_unsupportedConfigKey(t *testing.T) {
	networkName := acctest.GenerateName(2, "-")

	resource.Test(t, resource.TestCase{
		PreCheck:                 func() { acctest.PreCheck(t) },
		ProtoV6ProviderFactories: acctest.ProtoV6ProviderFactories,
		Steps: []resource.TestStep{
			{
				// A NAME placeholder matches exactly one segment.
				Config: acctest.Provider() + testAccNetwork_memberOverrides(networkName, "bridge", map[string]string{
					"bgp.peers.site.peer1.address": "192.0.2.1",
				}, nil),
				ExpectError: regexp.MustCompile(`does not support config key`),
			},
		},
	})
}
```

tugbataluy:
Nice one. This fixes a real papercut (`user.*` and `bgp.peers.NAME.*` were unusable on any server that exposes metadata) and the network/storage dedup into `common` is a clear win. The matcher rules line up with what LXD actually ships, and the tests read well. I gave two tiny comment nits and one optional negative test, but nothing is blocking, thanks @MusicDin :rocket: !

MusicDin:
Thanks @tugbataluy! Would you mind if I address your comments in a follow-up PR?

tugbataluy:
Sure of course :sparkle:

tugbataluy (approved):

MusicDin:
Comments addressed in https://github.com/terraform-lxd/terraform-provider-lxd/pull/781

PR #783, opened 2026-09-10
auth: Minor corrections on devlxd docs and error message

MusicDin:
Addresses remaining comments from https://github.com/terraform-lxd/terraform-provider-lxd/pull/771


tugbataluy (approved):
LGTM, thanks for addressing my concerns @MusicDin 🚀

